PR #8, opened 2016-11-24
Updating the matrix multiplication example

Ruyk:
- Host matrix multiplication deals with non-divisible sizes of matrix
- SYCL variant only accepts power-of-two matrix sizes
- Error correction is added to both host and sycl variants

### DIFF
--- a/samples/matrix_multiply/matrix_multiply.cpp
+++ b/samples/matrix_multiply/matrix_multiply.cpp
@@ -311,7 +311,6 @@ int main(int argc, char *argv[]) {
         }
       if (!error) {
         std::cout << "Success" << std::endl;
-        ;
       } else {
         std::cout << " Error in the computation " << std::endl;
       }
--- a/samples/matrix_multiply/matrix_multiply.cpp
+++ b/samples/matrix_multiply/matrix_multiply.cpp
@@ -62,10 +62,10 @@ void display_matrix(float *m, int matSize) {
 void block_host(float *MA, float *MB, float *MC, int matSize) {
   /* We set the block size to 32 for simplicity, though the optimal
    * value will depend on the platform this is run on. */
-  int block_size = 32; 
+  int block_size = 32;
   int numBlocks = block_size / matSize;
   int extraBlockLength = block_size % matSize;
-  numBlocks = extraBlockLength? (numBlocks + 1) : (numBlocks);
+  numBlocks = extraBlockLength ? (numBlocks + 1) : (numBlocks);
 
 #pragma omp parallel for collapse(2)
   for (int bIndexI = 0; bIndexI < matSize; bIndexI += block_size)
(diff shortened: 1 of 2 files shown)


samples/matrix_multiply/matrix_multiply.cpp
```diff
@@ -113,6 +135,9 @@ void local_mxm(cl::sycl::queue &q, T *MA, T *MB, T *MC, int matSize) {
             << std::endl;
   std::cout << " The order is : " << matSize << std::endl;
   std::cout << " The blockSize is : " << blockSize << std::endl;
+  // Make sure the block size is not larger than the mat size
+  blockSize = min(matSize, blockSize);
```
DuncanMcBain: This will be cl::sycl::min because of the using statement in this file

samples/matrix_multiply/matrix_multiply.cpp
```diff
@@ -271,65 +298,90 @@ int main(int argc, char *argv[]) {
...
+          if (std::fabs(MC[i * matSize + j] - MB[i * matSize + j]) > 1e-8) {
+            std::cout << " Position " << i << ", " << j
+              << " differs: " << MC[i * matSize + j]
+              << " != " << MB[i * matSize + j] << std::endl;
+            error = true;
+          }
+        }
+      if (!error) {
+        std::cout << "Success" << std::endl;
+        ;
```
DuncanMcBain: Lonely semi-colon

samples/matrix_multiply/matrix_multiply.cpp
```diff
@@ -271,65 +298,90 @@ int main(int argc, char *argv[]) {
...
+            end - start).count();
+        std::cout << "SYCL: ";
+        std::cout << "Time: " << time << std::endl;
+        float flops =
           (2.0f * matSize * matSize * matSize / (time / 1000.0f)) * 1.0e-9f;
-      std::cout << "GFLOPs: " << flops << std::endl;
-    }
-  }
+        std::cout << "GFLOPs: " << flops << std::endl;
+        std::cout << " Output " << std::endl;
```
DuncanMcBain: This will be printed on a line by itself

samples/matrix_multiply/matrix_multiply.cpp
```diff
@@ -271,65 +298,90 @@ int main(int argc, char *argv[]) {
...
-                  << " differs: " << MC[i * matSize + j]
-                  << " != " << MB[i * matSize + j] << std::endl;
-        error = true;
+      if (!error) {
+        display_matrix(MC, matSize);
+        error = false;
+        // Testing
+        for (int i = 0; i < matSize; i++)
+          for (int j = 0; j < matSize; j++) {
+            if (std::fabs(MC[i * matSize + j] - MB[i * matSize + j]) > 1e-8) {
```
DuncanMcBain: Can we replace this with something like the sample from here? Sounds pedantic, but I'm worried that for larger numbers (say ~100) that will be spuriously false, or that for small numbers it will be incorrectly true.

Ruyk: This was there before my latest changes, so I rather not change it in this patch.

samples/matrix_multiply/matrix_multiply.cpp
Ruyk: Removed

samples/matrix_multiply/matrix_multiply.cpp
DuncanMcBain: Oh, it was? My bad, I'll add it myself then, rather than be a hypocrite! But in a separate merge request. Thanks very much!

PR #11, opened 2017-01-04
Legacy pointer utility header

Ruyk:
This pull request contains a utility header that implements a malloc/free-like interface for buffers, on top of standard SYCL 1.2.
It includes two tests that serve as example.
This interface simplifies porting code from C-based GPU programming languages into SYCL.


utils/legacy_pointer/include/legacy_pointer.hpp
```diff
@@ -0,0 +1,259 @@
...
+   * Return the number of active pointers (i.e, pointers that
+   * have been malloc but not freed).
+   */
+  size_t count() const { return __pointer_list.size(); }
+
+ private:
+  /* Maps the buffer id numbers to the actual buffer
+   * instances.
+    */
+  std::map<buffer_id, buffer_t> __pointer_list;
```
DuncanMcBain: If this is a map rather than a list, could we call it m_pointerMap? Since we're indexing with effectively uints, could we use an unordered map instead? I suppose it will never make much of a difference, given the size it is likely to be.

utils/legacy_pointer/tests/offset.cc
```diff
@@ -0,0 +1,175 @@
...
+        buffer_t b = legacy::getPointerMapper().get_buffer(bId);
+
+        size_t offset = legacy::getPointerMapper().get_offset(actPos);
+        ASSERT_EQ(offset, (i * SIZE + j) * sizeof(float));
+        
+        q.submit([b, offset, i, j, SIZE](cl::sycl::handler& h) mutable {
+            auto accB = b.get_access<sycl_acc_rw>(h);
+            h.single_task(kernel(accB, i, j, SIZE, offset));
+        });
+        q.wait_and_throw();
```
DuncanMcBain: The queue has no async handler so there will be no throw here - could change to a wait, or add a handler

utils/legacy_pointer/include/legacy_pointer.hpp
```diff
@@ -0,0 +1,259 @@
...
+   * the offset:
+   *
+   * |== MAX_BUFFERS ==|======== MAX_OFFSET ========|
+   * |   Buffer Id     |       Offset in buffer     |
+   * |=================|============================|
+   */
+  struct legacy_pointer_t {
+    /* Type for the pointers
+    */
+    base_ptr_t _contents;
```
Ruyk: I normally would agree but in this case, each util may be coming from different projects (externals), and they may have their own indentation/coding scheme. Can we be more lenient for utils?

utils/legacy_pointer/include/legacy_pointer.hpp
```diff
@@ -0,0 +1,259 @@
...
+    //  buffer_t retVal = __pointer_list.at(bId);
+    auto it = __pointer_list.find(bId);
+    if (it != __pointer_list.end()) {
+      return it->second;
+    }
+
+    std::cerr << "No sycl buffer has been found. Make sure that you have "
+                 "allocated memory for your buffer by calling malloc function."
+              << std::endl;
+    std::abort();
```
Ruyk: Although I would normally agree with you, the project where this is used does not use exceptions, and they want things to just abort (not my decision!).

utils/legacy_pointer/include/legacy_pointer.hpp
Ruyk: Never mind, I'll fix it.

utils/legacy_pointer/include/legacy_pointer.hpp
Ruyk: Changing it to m_pointerMap.
As for unordered vs ordered, at this point, I'll stick with the map since it is already included in the projects we use it. We may change this in the future.

utils/legacy_pointer/tests/offset.cc
Ruyk: The wait and throw should not be required, will be removed.

utils/legacy_pointer/include/legacy_pointer.hpp
DuncanMcBain: OK, that makes sense. Thanks!

PR #17, opened 2017-01-26
CMake target building from multiple TUs

### LINKED ISSUE #15: CMake script supporting multiple translation units
The current version of [`add_sycl_to_target`](https://github.com/mcopik/computecpp-sdk/blob/master/cmake/Modules/FindComputeCpp.cmake#L238) supports only targets with one source file. It is a non-negligible limitation, many C++ applications or libraries consist of multiple translation units. Is there a technical reason behind this limitation?

In my fork I implemented [a single extension](https://github.com/mcopik/computecpp-sdk/blob/cmake_multiple_sources/cmake/Modules/FindComputeCpp.cmake#L246) creating multiple custom targets for each source file. I'm using an additional counter to distinguish between targets created for source files with the same name; it may happen for files located in different directories.
A [modified example](https://github.com/mcopik/computecpp-sdk/tree/cmake_multiple_sources_example/samples/example_sycl_application) proves that an application with SYCL kernels in two source files builds correctly; I can't verify if it executes correctly though, due to problems with SYCL runtime, but I do not expect any problems here.

I can open a pull request, as long as you don't see a problem with this approach and you are interested in it.

mcopik:
Fix #15
Supporting targets with multiple source files is a must-have feature for many projects.

### DIFF
--- a/cmake/Modules/FindComputeCpp.cmake
+++ b/cmake/Modules/FindComputeCpp.cmake
@@ -163,8 +163,11 @@ endif()
 #  targetName : Name of the target.
 #  sourceFile : Source file to be compiled.
 #  binaryDir : Intermediate directory to output the integration header.
+#  fileCounter : Counter included in name of custom target. Different counter
+#       values prevent duplicated names of custom target when source files with the same name,
+#       but located in different directories, are used for the same target.
 #
-function(__build_spir targetName sourceFile binaryDir)
+function(__build_spir targetName sourceFile binaryDir fileCounter)
   
   # Retrieve source file name.
   get_filename_component(sourceFileName ${sourceFile} NAME)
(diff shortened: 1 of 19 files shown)


cmake/Modules/FindComputeCpp.cmake
```diff
@@ -232,13 +238,18 @@ endfunction()
 #  target and sets a dependancy on that new command.
 #
 #  targetName : Name of the target to add a SYCL to.
-#  sourceFile : Source file to be compiled for SYCL.
 #  binaryDir : Intermediate directory to output the integration header.
+#  sourceFiles : Source files to be compiled for SYCL.
 #
-function(add_sycl_to_target targetName sourceFile binaryDir)
+function(add_sycl_to_target targetName binaryDir sourceFiles)
```
DuncanMcBain: This interface change breaks all the samples - is it necessary? Does "sourceFiles" represent a variable number of arguments that need to be at the end of the arguments to the function after this change?

DuncanMcBain:
Hi, this change looks really good apart from the one comment I've left about changing the interface to "add_sycl_to_target".

cmake/Modules/FindComputeCpp.cmake
mcopik: I forgot to discuss this change. I don't think that one can pass an undefined number of source files without placing them at the very end, unless we explicitly group arguments in the call, i.e.:
https://cmake.org/cmake/help/v3.0/module/CMakeParseArguments.html
Then the order of arguments does not matter but this solution would break the compatibility as well.

If you see another way of implementing it, please share it with me. Otherwise, I should fix my branch by changing CMakeLists for examples.

DuncanMcBain: I don't have any other ideas, I suspected that it was the varargs thing causing the change. If you're happy to alter the other CMake files to fit the change I'd be very grateful, though I'd be happy to fix it up as part of the merge - whatever works for you! :) (My CMake-fu is not very strong)

mcopik: Done. Samples can be built correctly now.

Ruyk:
Thanks for this changes @mcopik! it may be useful for some people. However, I am concern about the change of the interface, as other projects (Parallel STL, GPU-STREAM, ...) use the previous syntax. Can we have this as a new function like "add_multiple_sycl_to_target" and keep the old interface for a single file?

mcopik:
@Ruyk Well, it's up to you make that decision - I have made a suggestion how to implement a missing feature. I don't really have a preference here as long as it works. I chose this syntax because it is a standard convention.

DuncanMcBain:
I've decided that the change to the interface is good, it makes sense. I also suspect that there won't be changes to this function particularly often, and we'll just have to make the other projects aware of this change.

PR #21, opened 2017-03-01
Add missing headers preventing compilation on Ubuntu 16.10

keryell:
`std::pow` and `std::iota` need to be declared before use


DuncanMcBain:
Thank you very much, Ronan!

PR #24, opened 2017-05-04
Added parsing of target include dirs

MathiasMagnus:
Because it is fairly common that people do not set include directories for the given source folder, but on a per-target basis, I added the ability for the build scripts to detect such target properties and add them to the device compiler invocation. (plus fixed indentation on one line.)


DuncanMcBain:
Thanks very much!

PR #35, opened 2017-07-21
Add Python script to extract IR from integration header

Vtec234:


tools/extract_ir/extract_ir.py
```diff
@@ -0,0 +1,140 @@
...
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+#  Codeplay's ComputeCpp SDK
+#
+#  extract_ir.py
+#
+#  Description:
+#    Script to extract SPIR or other IR from a SYCL integration header.
+#
```
Ruyk: Add @davors72 (Stephen McGroarty) as author

DuncanMcBain:
Could you format this according to PEP8? This should help: https://pypi.python.org/pypi/autopep8

PR #43, opened 2017-08-23
Add user-specified base address to pointer mapper

DuncanMcBain:
If the base address of the pointer mapper is 1, various tools will
complain about alignment, even though these are not "real" pointers.
Changing it to a user-specified variable, with a default highly-aligned
value, should fix this problem.

### DIFF
--- a/utils/vptr/include/virtual_ptr.hpp
+++ b/utils/vptr/include/virtual_ptr.hpp
@@ -34,6 +34,7 @@
 
 #include <queue>
 #include <set>
+#include <stdexcept>
 #include <unordered_map>
 
 namespace cl {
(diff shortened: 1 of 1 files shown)


utils/vptr/include/virtual_ptr.hpp
```diff
@@ -267,7 +267,12 @@ class PointerMapper {
   /**
    * Constructs the PointerMapper structure.
    */
-  PointerMapper() : m_pointerMap{}, m_freeList{} {};
+  PointerMapper(base_ptr_t baseAddress = 4096)
+      : m_pointerMap{}, m_freeList{}, m_baseAddress{baseAddress} {
+    if (m_baseAddress == 0) {
+      throw std::out_of_range(std::string("Base address cannot be zero"));
```
Ruyk: I would say an std::invalid_argument is more relevant

DuncanMcBain: You're too right! I actually went onto cppreference, looked at the exceptions, thought to myself "this is invalid arg!" and wrote out of range because I'm a damn fool. Thanks!

PR #45, opened 2017-08-29
Improve inclusive scan to handle large arrays

Vtec234:


samples/scan/scan.cpp
```diff
@@ -57,104 +57,210 @@ struct identity<T, std::logical_and<T>> {
...
+ * on the data in the `in` buffer. Runs in parallel on the provided accelerated
+ * hardware queue. Modifies the input buffer to contain the results of the scan.
  * Input size has to be a power of two. If the size isn't so, the input can
  * easily be padded to the nearest power of two with any values, and the scan on
  * the meaningful part of the data will stay the same. */
 template <typename T, typename Op>
-std::vector<T> par_scan(std::vector<T> const& in, sycl::queue& q) {
-  size_t size = in.size();
-  std::vector<T> out(size);
+void par_scan(sycl::buffer<T, 1>& in, sycl::queue& q) {
```
DuncanMcBain: Why the interface change - to avoid the costly copy?

samples/scan/scan.cpp
```diff
@@ -57,104 +57,210 @@ struct identity<T, std::logical_and<T>> {
...
+    auto data = in.template get_access<sycl::access::mode::read_write>(cgh);
+    sycl::accessor<T, 1, sycl::access::mode::read_write,
+                   sycl::access::target::local>
 
-  return out;
-}
+        temp(wgroup_size * 2, cgh);
+
+    // Use dummy struct as the unique kernel name.
+    cgh.parallel_for<kernel_name<T, Op, 0>>(
```
DuncanMcBain: Hmmm, you could do a fun little trick here by making this in itself a `typename` in `kernel_name`'s declaration, that way you could call it "class segmented_scan", then the next like "class store_largest" or something, to help with readability (this would also come up in debugging). Consider it minor, though - and it's slightly unpleasant

samples/scan/scan.cpp
```diff
@@ -57,104 +57,210 @@ struct identity<T, std::logical_and<T>> {
...
+          data[off_gid * 2] = Op{}(data[off_gid * 2], ends_scan[group]);
+          data[off_gid * 2 + 1] = Op{}(data[off_gid * 2 + 1], ends_scan[group]);
+        });
+  });
+}
 
 /* Tests the scan with an addition operation, which is its most common use.
  * Returns 0 if successful, a nonzero value otherwise. */
 int test_sum(sycl::queue& q) {
+  constexpr size_t SIZE = 512;
```
DuncanMcBain: I think you don't need to capitalize this - it's a local var, not a macro, as much as it is constexpr.

samples/scan/scan.cpp
```diff
@@ -181,12 +287,22 @@ int test_sum(sycl::queue& q) {
 /* Tests the scan with a multiply operation, which is a sequence of factorials.
  * Returns 0 if successful, a nonzero value otherwise. */
 int test_factorial(sycl::queue& q) {
+  // Anything above this size overflows the int64_t type
+  constexpr size_t SIZE = 16;
```
DuncanMcBain: Same deal - I don't think caps are needed.

samples/scan/scan.cpp
Vtec234: With the previous interface, the recursive invocation would have to make a host-side vector. This avoids that.

DuncanMcBain: Ah of course, it's recursive, I see now. Thanks!

samples/scan/scan.cpp
Vtec234: Compute++ disallows using nested classes/structs/etc. as kernel names, and a typedef still resolves to the `kernel_name` class, so I left it as is.

Issue #32, opened 2017-07-03
Support for find_package in CMake


Vtec234:
To uphold standard practice in the CMake community, it would be good to define a finder module for ComputeCpp compatible with `find_package`. This module would define `COMPUTECPP_INCLUDE_DIR`, `COMPUTECPP_LIBRARY`, etc. Due to the nature of compiling SYCL code with a device compiler, we need to have a set of helper functions that support adding SYCL targets easily. These can be defined in a second CMake module. However, it is currently not possible to obtain an integration header for manual inclusion easily, which makes some more complicated CMake build scripts more complicated than necessary. For this reason, we could provide an `add_sycl_integration_header` function, which would set a user-provided variable to the path of the integration header. `add_sycl_to_target` would then use this function internally, but also include the integraton header for the user.

pretyman:
> This module would define COMPUTECPP_INCLUDE_DIR, COMPUTECPP_LIBRARY

Please don't. CMake is a mess by nature, but let's not advocate outdated practices. We should use targets instead and transient dependencies through the target's interface.

`find_package` should create a target, (ideally in a namespace) `ComputeCpp::ComputeCpp`, with the include dirs and libraries in its interface.

I wrote ComputeCppConfig.cmake and other CMake files which are auto generated by CMake, to mimic as if ComputeCpp package was CMake installed. I used https://github.com/forexample/package-example as a base and added the cmake files to the `lib/cmake/ComputeCpp`
```
find_package(ComputeCpp)
add_executable("my_executable" "src/main_with_kernel.cpp")
target_link_libraries("my_executable" PUBLIC ComputeCpp::ComputeCpp)
computecpp_compile("my_executable" "src/main_with_kernel.cpp")
```

In practice I ended up not using `computecpp-sdk`, I used it only as a base to understand what was going on in `FindComputeCpp.cmake`. This cmake file is a bit odd and seems wrong in parts, for example, it adds `-include file_with_kernel.cpp.sycl` file to the CMake target, but I guess you want to use `-include file_with_kernel.cpp.sycl` when compiling `file_with_kernel.cpp`, isnt that correct? Not to mention that if more than one source is passed to `add_sycl_to_target`, only the last one will have its  `-include` flag added to the target.

However there are things I have not yet fixed in `add_sycl_to_target`, like passing the include directory to computecpp, how it is today won't work 100% because the include directory property of CMake can contain the dreaded generator expressions, so the code today will eventually break when used with other projects.

[...]

Issue #14, opened 2017-01-21
CMake integration for ComputeCpp


mcopik:
We have been very happy to use CMake scripts for building SYCL apps with ComputeCpp in our project. It simplifies integration with existing codebases significantly, but our experience suggests that the current approach is not perfect. I don't know the rationale behind this design but so far I have seen two ways of integrating CMake in open-source projects:

- CMake config file - in a third-party project user is required to pass only one install directory where CMake config file can be located; variables are read and used later to find headers, libraries, compilation and linking flags etc.
- CMake `find_package` script

The first option is obviously not applicable in this scenario because the interesting part is code for adding targets for building SYCL code. The second option is convenient as long as the script is freely available, either already distributed with CMake itself or shipped with the package itself - then integration is built on top of one user-defined installation directory. In the case of ComputeCpp, the location of the script is completely independent of ComputeCpp's current location. Therefore, a user of HPX has to either modify `CMAKE_MODULE_PATH` on the command line or pass two variables, one used to extend module path internally and one used to locate ComputeCpp. It's a very non-standard solution and quite inconvenient.

We could have included FindComputeCpp in our project and distributed it with HPX but it is not a preferred option - we would have to keep watching your repo to not miss updates and solve possible licensing issues, which may not be a problem, but we are always cautious when adding a third-party code to HPX which is not on Boost license.

Is there a specific reason why CMake scripts are distributed with SDK? We believe that it's perfectly reasonable to include it in ComputeCpp installation package and it would allow SYCL-based projects to configure build with just one user-defined path.

DuncanMcBain:
Hi Marcin, we understand what you're asking for, however, we separated the CMake from the ComputeCpp releases because having it as part of the SDK allows us to make changes independently of the main package releases. These may be infrequent which would limit how often we could update FindComputeCpp.cmake,  which we'd rather avoid.

In future we plan to have an installer of sorts which would automatically clone the SDK for you, to keep things more in sync. From what I can tell, the boost licence is a modified apache licence, so there shouldn't be any conflicts there particularly. At the moment I suppose you could clone the SDK inside the ComputeCpp install directory (provided you don't install in a root location, which isn't always possible - I install to $HOME).

MathiasMagnus:
On a related note, but arriving at it from a totally different angle:

@DuncanMcBain, what are your thoughts on having upstream CMake support for ComputeCpp?

[...]

PR #50, opened 2017-09-11
Added cxx language standard detection

MathiasMagnus:
Added the ability to detect the required cxx standard version on a per-target basis. This is done to allow building SYCL targets with a specific language requirement as simply as:
```CMake
cmake_minimum_required(VERSION 3.2.2)

project(SYCL_CXX14 LANGUAGES CXX)

# Include modules
set(CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake/Modules)
include(FindComputeCpp)

add_executable(generic src/hello_generic.cpp)

set_target_properties(generic PROPERTIES CXX_STANDARD 14
                                         CXX_STANDARD_REQUIRED ON)

add_sycl_to_target(generic ${CMAKE_CURRENT_BINARY_DIR}
                   ${CMAKE_CURRENT_SOURCE_DIR}/src/hello_generic.cpp)
```

_On a separate note, I do plan on overhauling the CMake detection scripts of both ComputeCpp and triSYCL to make them work better together. I'll open an issue to demo what I had in mind (before I burn my time for something nobody is interested in)._


MathiasMagnus:
I forgot to mention. Computecpp-sdk samples compile fine without specifying any language standard, as expected. Extra behavior kicks in only when user specifies a standard to be used.

DuncanMcBain:
Hi @MathiasMagnus, thanks for your patch! I'd be interested to know if this has come up in your development process, because realistically for ComputeCpp only two language versions make sense - 11 and 14. The options provided by computecpp_info will set the language version to 14, unless otherwise specified on the command line, but I can't think of any scenario where setting C++14 will break code that would otherwise work. Similarly we cannot in any way support C++98, and while clang-3.9 (on which our compiler is based) has decent support for C++ 17 there's no guarantee of it working in or with SYCL code.

I would be interested to know if you've managed to use any C++17 with our compiler. As I say it is possible that things might "just work", but they could just as easily fail, unfortunately.

That being said, do you also envision this being useful for stopping use of features that are from a later standard? e.g. accidentally using auto lambdas or similar when you the code is supposed to be C++11 only?

There is a separate issue (#32) for improving the CMake, and I have promised myself I'll make that change one day. Additionally there's issue #14 which I would like to revisit, as it probably does make sense for the CMake to be available in the package (as much as it's been useful for us to be able to update FindComputeCpp.cmake as and when necessary). Please feel free to get involved with those - it's an area that I've neglected, as you can see by the age of the issues, but I will make it better at some point.

Thanks,
Duncan.

MathiasMagnus:
We too generally opt-in for the highest possible language standard that is portable enough to not cause too many headaches. Currently that is C++14 (though MSVC is still lacking, last time I checked Boost::Hana still did not compile).

You are correct, it might be smarter to throw a hard error on C++98, even though no entry point of SYCL explicitly requires C++11. cl::sycl::queue::submit can make do with an ordinary functor, it isn't mandated to be a lambda. Still, the implementations of SYCL are certainly sprinkled with C++11 all over the place.

As for C++17, the only reason we do not rely on it just yet is because it is far from widely available, it is roughly half done in Clang 3.9, but you are right there are nice features already implemented. constexpr if, fold expressions (I am just about to bake something instead of it for C++14), inline variables, just to name the most useful ones. (constexpr lambdas and template argument deduction for class templates missing in 3.9, roughly the 2 missing C++17 features we'd wish to rely on.) We will surely play with these as well, though production code will still target C++14 for a few months.

I am aware of those two issues, and I very much wanted to shove every extra logic into an imported target, however that does not allow running scripts on the target to extract the info we do today with add_sycl_to_target. I was wondering whether ComputeCpp and triSYCL are mature enough to have actual CMake support, such that CUDA received in CMake 3.9. I don't see that either project had their compilation model baked to a point that "rigid" external tooling would be a good idea. Other than that, the scripts could still be improved.

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -193,6 +193,21 @@ function(__build_spir targetName sourceFile binaryDir fileCounter)
...
+  set(device_compiler_cxx_standard)
+  get_target_property(targetCxxStandard ${targetName} CXX_STANDARD)
+  if (targetCxxStandard MATCHES 17)
+    set(device_compiler_cxx_standard "-std=c++1z")
+  elseif (targetCxxStandard MATCHES 14)
+    set(device_compiler_cxx_standard "-std=c++14")
+  elseif (targetCxxStandard MATCHES 11)
+    set(device_compiler_cxx_standard "-std=c++11")
+  elseif (targetCxxStandard MATCHES 98)
+    message(FATAL_ERROR "SYCL implementations cannot be compiled using C++98")
```
keryell: What about C++03 and some other weird versions?
I guess we do not care... :-)

MathiasMagnus: These are the ones CMake knows.

https://cmake.org/cmake/help/v3.9/prop_tgt/CXX_STANDARD.html

I'm not discriminating anything here... ;)

Ruyk:
Does this work with older CMake versions? We are currently supporting 3.2.2 and above. If we have to bump the required version of CMake for the SDK we will have to adapt the documentation accordingly...

MathiasMagnus:
It does. I do not have an installation available, but the [online docs for 3.2](https://cmake.org/cmake/help/v3.2/prop_tgt/CXX_STANDARD.html?highlight=cxx_standard) already mention this variable. 17 as an option was added in 3.8, but there's really no need to check. If the query returns 17, the user is definitely running 3.8+. If the user is running pre-3.8 then the query can never yield the value 17 (because it cannot be set), but the script engine won't argue about a string check.

Ruyk:
Looks good to me then!

PR #51, opened 2017-09-18
Msvc compat

MathiasMagnus:
Having obtained the Windows installer from @rodburns of ComputeCpp, I went ahead and fixed the scripts to enable building on Windows using MSVC.

## Note
Currently only C++14 language mode is supported, because compute++.exe fails to parse the STL headers, as some contain auto return type deduction, which is a C++14 feature. Most likely compute++.exe does not define certain things that triggers alternate STL code paths.

I would like to have these script modifications merged, so people can make use of it at as soon as possible. I will try to get around to detect the actual MSVC version as well as platform toolset used and emit warnings as necessary.

Currently I used best and latest VS 15.4 Preview 2 with platform toolset v141.


cmake/Modules/FindComputeCpp.cmake
```diff
@@ -242,7 +256,11 @@ function(__build_spir targetName sourceFile binaryDir fileCounter)
   add_dependencies(${targetName} ${headerTargetName})
 
   # Force inclusion of the integration header for the host compiler
-  set(compileFlags -include ${outputSyclFile})
+  if(MSVC)
+    set(compileFlags /FI ${outputSyclFile})
```
DuncanMcBain: Since this is only used to propagate the force include, maybe now would be a good time to rename it to forceIncludeFlag or similar? Since that is its only purpose.

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -275,12 +293,14 @@ function(add_sycl_to_target targetName binaryDir sourceFiles)
   )
   # Add custom target to run compute++ and generate the integration header
   foreach(sourceFile ${sourceFiles})
+    message (STATUS "binaryDir = ${binaryDir}")
```
DuncanMcBain: Was this a debugging print?

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -275,12 +293,14 @@ function(add_sycl_to_target targetName binaryDir sourceFiles)
...
     __build_spir(${targetName} ${sourceFile} ${binaryDir} ${fileCounter})
     MATH(EXPR fileCounter "${fileCounter} + 1")
   endforeach()
 
   # Link with the ComputeCpp runtime library
-  target_link_libraries(${targetName} PUBLIC ${COMPUTECPP_RUNTIME_LIBRARY}
-                        PUBLIC ${OpenCL_LIBRARIES})
+  target_link_libraries(${targetName} PUBLIC $<$<OR:$<CONFIG:Debug>,$<CONFIG:RelWithDebInfo>>:${COMPUTECPP_RUNTIME_LIBRARY_DEBUG}>
```
DuncanMcBain: I realise this solves the problem rather neatly but I'd prefer for the logic to be factored out, put into a different variable and then referenced in the link_libraries call - this isn't very readable!

Ruyk: I am not sure it can be factored out due to the nature of generator expressions in CMake...

DuncanMcBain: OK, that's fair. I guess what I'm saying is: I won't ever be able to change that, probably :smiley:

cmake/Modules/FindComputeCpp.cmake
MathiasMagnus: ok, will change.

cmake/Modules/FindComputeCpp.cmake
MathiasMagnus: My bad. Remaining debug message. Will remove.

cmake/Modules/FindComputeCpp.cmake
MathiasMagnus: It can be changed to ordinary script logic at the cost of LoC. I wanted it to remain concise. I do agree the generator expression syntax is not the nicest, but it is less LoC. Will change it to `if-else` if you wish.

DuncanMcBain: If it can't be moved, fair enough. In that case just leave it. It's unlikely to need to change, so I'll just have to learn to love it!

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -50,8 +50,8 @@ elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
       message(STATUS "host compiler - clang ${CMAKE_CXX_COMPILER_VERSION}")
     endif()
 else()
-  message(WARNING
-    "host compiler - Not found! (ComputeCpp supports GCC and Clang, see readme)")
+  set (COMPUTECPP_PLATFORM_SPECIFIC_ARGS "-fno-ms-compatibility")
+  #message(WARNING "host compiler - Not found! (ComputeCpp supports GCC and Clang, see readme)")
```
DuncanMcBain: This can be removed. Arguably this whole if-else chain could be removed, as other compilers are unlikely to work (though maybe icc would? And if it is capable, why would we stop it?)

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -50,8 +50,8 @@ elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
       message(STATUS "host compiler - clang ${CMAKE_CXX_COMPILER_VERSION}")
     endif()
 else()
-  message(WARNING
-    "host compiler - Not found! (ComputeCpp supports GCC and Clang, see readme)")
+  set (COMPUTECPP_PLATFORM_SPECIFIC_ARGS "-fno-ms-compatibility")
```
DuncanMcBain: Is this a clang-only flag? Would it make sure that MSVC-isms don't sneak into the SDK?

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -97,15 +97,29 @@ else()
...
+find_library(COMPUTECPP_RUNTIME_LIBRARY_DEBUG
+  NAMES ComputeCpp ComputeCpp_vs2015_d
+  PATHS ${COMPUTECPP_PACKAGE_ROOT_DIR}
+  HINTS ${COMPUTECPP_PACKAGE_ROOT_DIR}/lib PATH_SUFFIXES lib
+  DOC "ComputeCpp Debug Runtime Library" NO_DEFAULT_PATH)
+if (EXISTS ${COMPUTECPP_RUNTIME_LIBRARY_DEBUG})
+  mark_as_advanced(COMPUTECPP_RUNTIME_LIBRARY_DEBUG)
+  message(STATUS "${COMPUTECPP_RUNTIME_LIBRARY_DEBUG} - Found")
+else()
+message(FATAL_ERROR "ComputeCpp Debug Runtime Library - Not found!")
```
DuncanMcBain: should this be indented? If I understand this correctly, this is saying that for non-Windows systems, this will look for the ordinary library and set the debug variable to that as well?

DuncanMcBain:
Thanks for this, we hadn't added support for Windows to the SDK because the ComputeCpp package isn't available to Windows yet, but this is good so when you mark those comments as resolved I'll merge this in :+1:

cmake/Modules/FindComputeCpp.cmake
MathiasMagnus: This is a workaround for a Clang bug there is no fix to just yet.

http://clang-developers.42468.n3.nabble.com/Why-in-C-11-target-win32-char16-t-disappears-td4034906.html

The problem is, that strangely Clang when put into C++11/14 mode does not define the fundamental types: char16_t and char32_t which plague the STL all over. If no-ms-compat is passed onto the compiler, it does. Hence when MSVC hosts Clangs STL, it is required to put into strict C++ mode; this flag turns off some MSVC weirdness. Luckily, MSVC is becoming conforming enough that the STL compiles fine in this mode.

cmake/Modules/FindComputeCpp.cmake
MathiasMagnus: It should be. Indeed, it does just that. I wanted to minimize to divergent script flow control. I wanted to make the script generally less verbose. At some later point in time I wanted to make the scripts (optionally at least) a bit more silent. Developers don't usually care about things like stock compute++.exe flags.

DuncanMcBain: That's fair. I'm not particularly bothered about the output from the script, if at some stage you wanted to remove some of the noisier messages I'd be down to remove them.

cmake/Modules/FindComputeCpp.cmake
DuncanMcBain: Bizarre! Fair enough, whatever makes it work.

DuncanMcBain:
Ah yes, I meant to respond to the version stuff: I'm not worried about C++ version support on Windows, if the platform doesn't really let you choose what version you're targeting then I don't think we need to be fussy either. I guess the ideal would be that the SDK works "out of the box", if users somehow want more complicated configurations then they can build on this, but the base is a just-works solution. If that makes sense.

MathiasMagnus:
I did not quite understand the version selection part. MSVC does let you select language standards through the `/std:c++14`, `/std:c++17` and `/std:c++latest` compiler switches, and CMake does just this through the `CXX_STANDARD` target property. The trouble I believe is due to Clang not defining something that MSVC does when a specific standard is selected, and parts of the STL are not `#ifdef`-ed correctly.

_The only violation of a language feature being 'backported', is if-constexpr being available in C++14 mode for MSVC, because the STL implementers really wanted to use the metaprogramming simplifications if-constexpr enables within the implementation of the STL._

DuncanMcBain:
Shows how much I know then! I'd like to merge this, I think that's everything sorted. Thank you very much for doing this, that's a great improvement!

PR #53, opened 2017-09-27
added windows sdk support

Cp-Graham:


Cp-Graham:
existing implementation of cmake solution is being used by public currently and seems adequate.

Issue #40, opened 2017-08-05
Travis scripts rely on hardcoded ComputeCpp version number


DuncanMcBain:
At the moment, the travis config will fail when we update ComputeCpp as the version numbers change. We should change the scripts so that either they always download the same version (so we can update the number and avoid a failing config) or so that they will use whatever version is downloaded.

The travis config is broken at the moment, I'll fix it for now but a longer-term fix would do something like the above.

DuncanMcBain:
Fixed by !56.

PR #56, opened 2017-09-27
Update travis config for 0.3.2

DuncanMcBain:
Hopefully, this config will also not break when a new version is
released, as it now specifies the version of ComputeCpp to download.


DuncanMcBain:
Fixes bug #40.

PR #58, opened 2017-10-04
Add option whether to force include sycl file

Zakor94:
The sycl file should not be included this way when using Eigen for instance because it has to be included after the source file.
When compiling with CMake I have to create a temporary source file that includes the cpp file first and then the sycl file. A cleaner solution would be to compile the sycl file (with -xc++) and to include the original source file but I haven't been able to make this work with CMake.


DuncanMcBain:
This has been improved in 411ec07, which allows the user to select whether to include before or after the main code using a property scoped to the target level (which in FindComputeCpp.cmake's terms is equivalent to a single source file). Many thanks to @Zakor94 for the idea of how to actually do this!

PR #59, opened 2017-10-09
Add SPIR-V support to IR extraction utility

alistairlow:
I have tested this with SPIR-V and spirv-dis, but not with SPIR and llvm-dis. If the reviewer has some SPIR and llvm-dis handy, I would appreciate their help in testing!

### DIFF
--- a/tools/extract_ir/extract_ir.py
+++ b/tools/extract_ir/extract_ir.py
@@ -24,6 +24,7 @@
 #    Script to extract SPIR or other IR from a SYCL integration header.
 #
 #  Authors:
+#    Alistair Low <alistair@codeplay.com>
 #    Stephen McGroarty <stephen@codeplay.com>
 #    Wojciech Nawrocki <wojciech.nawrocki@codeplay.com>
 #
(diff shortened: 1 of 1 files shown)


DuncanMcBain:
Thanks for this. I have llvm-dis and a bunch of headers around, I will try your branch and if it works probably just merge. Thanks!

tools/extract_ir/extract_ir.py
```diff
@@ -53,37 +56,39 @@ def parse_args():
...
-    Given to file names, tries to use llvm-dis to disassemble the input file into the output file.
-    The output file defaults to stdout.
+    Given two file names, uses a disassembler to disassemble the input file into
+    the output file.
     """
     code = subprocess.call(
-        [llvm_dis_path,
-         input_file_name,
-         '-o=' + output_file_name])
+        [disassembler_path, '-o ' + output_file_name, input_file_name])
```
DuncanMcBain: If you pass "-o" and the filename as one string like this, you end up with both in one element of argv (e.g. argv[2] == "-o -" which leads to "unknown command line option" errors from llvm-dis. If you instead change it to "-o", "-", it'll work.

PR #61, opened 2017-10-23
Add a flag to enable SPIR-V generation

Ruyk:
COMPUTECPP_USE_SPIRV CMake flag can be used to force FindComputeCpp
module to generate SPIR-V binaries instead of SPIR.
The default behavour is to output SPIR binaries.


DuncanMcBain:
Manual merge: c9bdc56326f7ec5975104fd41eddb33ba0705d2d

PR #63, opened 2017-11-13
CMakeLists.txt - minimum version before project command and append to CMAKE_MODULE_PATH

pretyman:
Hi,

I updated the CMakeLists.txt to change the order of the project command As specified at the documentation: https://cmake.org/cmake/help/latest/command/project.html

And I also updated the overwrite of CMAKE_MODULE_PATH, which is better to append, so you don't lose the current values.

Thanks

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.2.2)
 
 project(ComputeCpp-SDK)
 
-set(CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/Modules)
+list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/Modules)
 include(FindComputeCpp)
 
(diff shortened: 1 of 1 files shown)


DuncanMcBain:
Hi @pretyman, thanks very much for your contribution! I didn't know that CMake recommended setting the minimum first, so I've learned something today! Again I never felt good about overwriting the module path, this seems much better, so many thanks.

I'm very happy to accept this, and will do so, but was wondering if you could sign the CLA we have set up through clahub - we wanted to provide certain guarantees to the folks who contributed patches, and this is the way in which we do that. I don't think there's anything dodgy in it - it is based very strongly on the original Apache CLA - but if there's a problem let me know.

You can sort this via the "details" button next to the failed check, next to the Travis build status (which, pleasingly, I see has passed).

Thanks again,
Duncan.

pretyman:
Thanks, done.

Sorry for taking long, this got buried with other github notices and I only came back to see today :-)

DuncanMcBain:
That's no problem! Thanks very much for this, I'll merge it now :)

PR #64, opened 2017-11-14
Allow to add existing buffer

Rbiessy:
This is a change we have in Eigen which allow to use the library with existing SYCL buffers.
So one can write
```
buffer<float, 1> buf(range<1>(10));
float* p = static_cast<float *>(static_cast<void *>(pMap.add_pointer(buf)));
```
and use p as if it were malloc'ed.

I'd like to make sure that this won't cause any issue.

### DIFF
--- a/include/vptr/virtual_ptr.hpp
+++ b/include/vptr/virtual_ptr.hpp
@@ -316,48 +316,17 @@ class PointerMapper {
   }
 
   /* add_pointer.
+   * Adds an existing pointer to the map and returns the virtual pointer id.
+   */
+  inline virtual_pointer_t add_pointer(buffer_t &b) {
+    return add_pointer_impl(b);
+  }
+
+  /* add_pointer.
    * Adds a pointer to the map and returns the virtual pointer id.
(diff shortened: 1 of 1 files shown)


include/vptr/virtual_ptr.hpp
```diff
@@ -316,48 +316,17 @@ class PointerMapper {
...
-
-        // update size of the current node
-        lastElemIter->second.m_size = bufSize;
-
-        // add the new free node
-        auto newFreePtr = lastElemIter->first + bufSize;
-        auto freeNode = m_pointerMap.emplace(newFreePtr, p2).first;
-        m_freeList.emplace(freeNode);
-      }
+  inline virtual_pointer_t add_pointer(buffer_t &b) {
```
Ruyk: why is this passed by reference? is the buffer going to be modified? If the only thing intended is to create a copy seems to me its better to pass by const &

Zakor94: Indeed good point

PR #65, opened 2017-11-14
Updated vptr README

Rbiessy:
I don't think the example was meant to be compiled but I guess it's better if it does compile :)


include/vptr/README.adoc
```diff
@@ -34,7 +34,7 @@ float * c = a + 5;
 assert(pMap.get_buffer(a) == pMap.get_buffer(c))
```
Rbiessy: I couldn't compile this line either because of the lack of operator== for buffers.

DuncanMcBain: Hmm, if the code doesn't work people might try it only to have it fail. That said, nobody has complained so far...

PR #67, opened 2017-11-30
Add method release_pointer to vptr

Rbiessy:
The idea is to allow the user to remove pointers without reusing them.
It is particularly useful for sub-buffers because we don't want to keep their virtual pointer. With that it is possible to use Eigen with sub-buffers.

I am not too sure about the method name. Maybe it would make more sense to swap it with the existing remove_pointer (i.e. release_pointer allow to re-use the pointer but remove_pointer does not)?. I'll make sure Eigen synchronises with the latest changes we make here.

### DIFF
--- a/include/vptr/virtual_ptr.hpp
+++ b/include/vptr/virtual_ptr.hpp
@@ -378,7 +378,9 @@ class PointerMapper {
 
   /* remove_pointer.
    * Removes the given pointer from the map.
+   * The pointer is allowed to be reused only if ReUse if true.
    */
+  template <bool ReUse = true>
   void remove_pointer(const virtual_pointer_t ptr) {
     auto node = this->get_node(ptr);
 
@@ -478,6 +480,15 @@ class PointerMapper {
   size_t m_baseAddress;
 };
 
+/* remove_pointer.
+ * Removes the given pointer from the map.
+ * The pointer is allowed to be reused only if ReUse if true.
+ */
+template <>
+void PointerMapper::remove_pointer<false>(const virtual_pointer_t ptr) {
+  m_pointerMap.erase(this->get_node(ptr));
+}
+
 /**
  * Malloc-like interface to the pointer-mapper.
  * Given a size, creates a byte-typed buffer and returns a
@@ -498,10 +509,12 @@ inline void *SYCLmalloc(size_t size, PointerMapper &pMap) {
  * Free-like interface to the pointer mapper.
  * Given a fake-pointer created with the virtual-pointer malloc,
  * destroys the buffer and remove it from the list.
+ * If ReUse is false, the pointer is not added to the freeList,
+ * it should be false only for sub-buffers.
  */
-template <typename PointerMapper>
+template <bool ReUse = true, typename PointerMapper>
 inline void SYCLfree(void *ptr, PointerMapper &pMap) {
-  pMap.remove_pointer(ptr);
+  pMap.template remove_pointer<ReUse>(ptr);
 }
 
 /**
--- a/include/vptr/virtual_ptr.hpp
+++ b/include/vptr/virtual_ptr.hpp
@@ -485,7 +485,7 @@ class PointerMapper {
  * The pointer is allowed to be reused only if ReUse if true.
  */
 template <>
-void PointerMapper::remove_pointer<false>(const virtual_pointer_t ptr) {
+inline void PointerMapper::remove_pointer<false>(const virtual_pointer_t ptr) {
   m_pointerMap.erase(this->get_node(ptr));
 }
 
--- a/tests/vptr/basic.cc
+++ b/tests/vptr/basic.cc
@@ -167,6 +167,13 @@ TEST(pointer_mapper, reuse_ptr) {
     ASSERT_FALSE(PointerMapper::is_nullptr(shouldBeTheSame));
     ASSERT_EQ(pMap.count(), 3u);
     ASSERT_EQ(shouldBeTheSame, reused);
+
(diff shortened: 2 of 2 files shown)


include/vptr/virtual_ptr.hpp
```diff
@@ -398,6 +398,13 @@ class PointerMapper {
     }
   }
 
+  /* release_pointer.
+   * Removes the given pointer from the map and do not try to re-use it later.
+   */
+  void release_pointer(const virtual_pointer_t ptr) {
+     m_pointerMap.erase(this->get_node(ptr));
+  }
```
Ruyk: Is not clear to me the difference between release and remove. In fact, for a release I would expect that the buffer is released from the pointer map and returned to the user for their own purposes (like std::unique_ptr::release).
Wouldn't it be better if  we put this functionality on the existing remove_pointer and simply pass a flag to it to not re-use?

Rbiessy: Hum so I don't need the exact same thing as the unique_pre::release. That's because I assume the user already has the buffer.
Anyway I'm fine with adding a flag to remove_pointer to not re-use it. Are you ok with a templated bool parameter? Or do we want a runtime flag for some reason? I think a runtime flag would make sense if we could know whether a buffer is a sub-buffer at runtime but I don't think it's possible. So I would suggest for the template parameter.

Rbiessy: About the tests that's a fair point. I could work on adding a simple one tomorrow maybe.

Rbiessy:
I updated the PR with a template bool parameter for remove_pointer instead of release_pointer.
I added 2 more tests, one for the SYCLFree<false> and one for adding an existing pointer. Unfortunately I couldn't reproduce an interesting case with sub-buffers where reusing them causes a crash but I think it makes sense that we don't want to reuse them.

tests/vptr/basic.cc
```diff
@@ -170,6 +170,78 @@ TEST(pointer_mapper, reuse_ptr) {
...
+    ASSERT_FALSE(PointerMapper::is_nullptr(not_reused));
+    ASSERT_EQ(pMap.count(), 2u);
+
+    // Another large buffer
+    void *end = SYCLmalloc(100 * sizeof(int), pMap);
+    ASSERT_NE(end, nullptr);
+    ASSERT_FALSE(PointerMapper::is_nullptr(end));
+    ASSERT_EQ(pMap.count(), 3u);
+
+    // We free the intermediate one and forbit it to be reuse
```
Ruyk: forbit -> forbid

tests/vptr/basic.cc
```diff
@@ -170,6 +170,78 @@ TEST(pointer_mapper, reuse_ptr) {
...
+    SYCLfree<false>(not_reused, pMap);
+    ASSERT_EQ(pMap.count(), 2u);
+
+    void *shouldBeNew = SYCLmalloc(10 * sizeof(int), pMap);
+    ASSERT_NE(shouldBeNew, nullptr);
+    ASSERT_FALSE(PointerMapper::is_nullptr(shouldBeNew));
+    ASSERT_EQ(pMap.count(), 3u);
+    ASSERT_NE(shouldBeNew, not_reused);
+  }
```
Ruyk: Could you finish the test by freeing everything that was allocated an ensuring the pMap.count() is back to 0 ?

tests/vptr/basic.cc
```diff
@@ -170,6 +170,78 @@ TEST(pointer_mapper, reuse_ptr) {
...
+    // We free the intermediate one
+    SYCLfree(reused, pMap);
+    ASSERT_EQ(pMap.count(), 2u);
+
+    void *shouldBeTheSame = SYCLmalloc(10 * sizeof(int), pMap);
+    ASSERT_NE(shouldBeTheSame, nullptr);
+    ASSERT_FALSE(PointerMapper::is_nullptr(shouldBeTheSame));
+    ASSERT_EQ(pMap.count(), 3u);
+    ASSERT_EQ(shouldBeTheSame, reused);
+  }
```
Ruyk: Could you finish the test by freeing everything that was allocated an ensuring the pMap.count() is back to 0 ?

tests/vptr/basic.cc
Rbiessy: Oops. Maybe "reused" also?

Ruyk: indeed

tests/vptr/basic.cc
```diff
@@ -167,6 +167,99 @@ TEST(pointer_mapper, reuse_ptr) {
     ASSERT_FALSE(PointerMapper::is_nullptr(shouldBeTheSame));
     ASSERT_EQ(pMap.count(), 3u);
     ASSERT_EQ(shouldBeTheSame, reused);
+
+    SYCLfree(shouldBeTheSame, pMap);
+    ASSERT_EQ(pMap.count(), 2u);
+    SYCLfree(end, pMap);
+    ASSERT_EQ(pMap.count(), 1u);
+    SYCLfree(initial, pMap);
+    ASSERT_EQ(pMap.count(), 0u);
```
Rbiessy: My tests are based on reuse_ptr, for symmetry I also free'd everything here.

Ruyk: ok, makes sense

Rbiessy:
Can I squash and merge this?

Ruyk:
Go ahead

Rbiessy:
Thanks! I realised too late the name of the commit isn't great, I hope it's not a big deal...

Issue #69, opened 2017-12-06
PTX support


pkestene:
I noticed that ComputeCpp 0.5.0 has PTX support.

FindComputeCpp.cmake should probably be modified to enable option "--use-ptx"

DuncanMcBain:
Implemented in #70, thanks very much!

PR #70, opened 2017-12-06
minimal modification to make samples work on Nvidia gpus

pkestene:
This is a minimalist modification to enable building samples for Nvidia gpus (see issue #69 ).

To configure:
```bash
cmake -DCOMPUTECPP_USE_PTX=1 ..
```

### DIFF
--- a/cmake/Modules/FindComputeCpp.cmake
+++ b/cmake/Modules/FindComputeCpp.cmake
@@ -159,8 +159,13 @@ if (COMPUTECPP_USE_SPIRV)
   set(USE_SPIRV "--use-spirv")
 endif()
 
+set(USE_PTX "")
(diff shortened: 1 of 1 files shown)


DuncanMcBain:
Hi @pkestene, thanks for doing this! I'd been meaning to add this very functionality to the SDK but hadn't had a chance to fix it. This looks perfect, so I would like to merge it - however we have a CLA we like contributors to sign. It looks like CLAHub was having some problems earlier, but I've retried and if you look you can see the unsuccessful check above. If you have a read of that, we can then get this merged.
As you have likely guessed, the PTX support is in its early days. How many tests are you managing to run successfully?

DuncanMcBain:
Thanks very much for this!

PR #76, opened 2017-12-15
Add include-after support to MSVC

DuncanMcBain:
Previously the method of including the integration header after the user
source only worked for clang/gcc. It is possible to use a similar flag
on Windows, /TP, which indicates that all files should be compiled as
C++.

### DIFF
--- a/cmake/Modules/FindComputeCpp.cmake
+++ b/cmake/Modules/FindComputeCpp.cmake
@@ -292,33 +292,36 @@ function(__build_spir targetName sourceFile binaryDir fileCounter)
   # Add a dependency on the integration header
   add_dependencies(${targetName} ${headerTargetName})
 
+  # This property can be set on a per-target basis to indicate that the
+  # integration header should appear after the main source listing
+  get_property(includeAfter TARGET ${targetName}
+      PROPERTY COMPUTECPP_INCLUDE_AFTER)
+
+  if(includeAfter)
+    # Change the source file to the integration header - e.g.
+    # g++ -c source_file_name.cpp.sycl
+    set_property(TARGET ${targetName} PROPERTY SOURCES ${outputSyclFile})
+    # CMake/gcc don't know what language a .sycl file is, so tell them
+    set_property(SOURCE ${outputSyclFile} PROPERTY LANGUAGE CXX)
+    set(includedFile ${sourceFile})
+  else()
+    set(includedFile ${outputSyclFile})
+  endif()
+
   # Force inclusion of the integration header for the host compiler
   if(MSVC)
     # NOTE: The Visual Studio generators parse compile flags differently,
     # hence the different argument syntax
     if(CMAKE_GENERATOR MATCHES "Visual Studio")
-      set(forceIncludeFlags "/FI\"${outputSyclFile}\"")
+      set(forceIncludeFlags "/FI\"${includedFile}\" /TP")
     else()
-      set(forceIncludeFlags /FI ${outputSyclFile})
+      set(forceIncludeFlags /FI ${includedFile} /TP)
     endif()
   else()
-    # This property can be set on a per-target basis to indicate that the
-    # integration header should appear after the main source listing
-    get_property(includeAfter TARGET ${targetName}
-        PROPERTY COMPUTECPP_INCLUDE_AFTER)
-    if(includeAfter)
-      # Change the source file to the integration header - i.e.
-      # g++ -c source_file_name.cpp.sycl
-      set_property(TARGET ${targetName} PROPERTY SOURCES ${outputSyclFile})
-      # CMake/gcc don't know what language a .sycl file is, so tell them
-      set_property(SOURCE ${outputSyclFile} PROPERTY LANGUAGE CXX)
-      set(forceIncludeFlags "-include ${sourceFile} -x c++")
-    else()
-      set(forceIncludeFlags "-include ${outputSyclFile}")
-    endif()
+      set(forceIncludeFlags "-include ${includedFile} -x c++")
   endif()
-  # Seems that target_compile_options remove duplicated flags, which
-  # does not work for -include which cannot be removed (CMake bug #15826)
+  # target_compile_options removes duplicated flags, which does not work
+  # for -include (it should appear once per included file - CMake bug #15826)
   #   target_compile_options(${targetName} BEFORE PUBLIC ${forceIncludeFlags})
   # To avoid the problem, we get the value of the property and manually append
   # it to the previous status.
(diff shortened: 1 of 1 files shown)


cmake/Modules/FindComputeCpp.cmake
```diff
@@ -292,33 +292,36 @@ function(__build_spir targetName sourceFile binaryDir fileCounter)
...
+  # integration header should appear after the main source listing
+  get_property(includeAfter TARGET ${targetName}
+      PROPERTY COMPUTECPP_INCLUDE_AFTER)
+
+  if(includeAfter)
+    # Change the source file to the integration header - e.g.
+    # g++ -c source_file_name.cpp.sycl
+    set_property(TARGET ${targetName} PROPERTY SOURCES ${outputSyclFile})
+    # CMake/gcc don't know what language a .sycl file is, so tell them
+    set_property(SOURCE ${outputSyclFile} PROPERTY LANGUAGE CXX)
```
Ruyk: you should set LANGUAGE_VERSION here too

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -292,33 +292,36 @@ function(__build_spir targetName sourceFile binaryDir fileCounter)
...
-      set(forceIncludeFlags "-include ${sourceFile} -x c++")
-    else()
-      set(forceIncludeFlags "-include ${outputSyclFile}")
-    endif()
+      set(forceIncludeFlags "-include ${includedFile} -x c++")
   endif()
-  # Seems that target_compile_options remove duplicated flags, which
-  # does not work for -include which cannot be removed (CMake bug #15826)
+  # target_compile_options removes duplicated flags, which does not work
+  # for -include (it should appear once per included file - CMake bug #15826)
```
Ruyk: Thanks for the improved English :smile:

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -292,33 +292,36 @@ function(__build_spir targetName sourceFile binaryDir fileCounter)
...
   # Force inclusion of the integration header for the host compiler
   if(MSVC)
     # NOTE: The Visual Studio generators parse compile flags differently,
     # hence the different argument syntax
     if(CMAKE_GENERATOR MATCHES "Visual Studio")
-      set(forceIncludeFlags "/FI\"${outputSyclFile}\"")
+      set(forceIncludeFlags "/FI\"${includedFile}\" /TP")
     else()
-      set(forceIncludeFlags /FI ${outputSyclFile})
+      set(forceIncludeFlags /FI ${includedFile} /TP)
```
Ruyk: What is this else for? ninja?

cmake/Modules/FindComputeCpp.cmake
DuncanMcBain: We don't set the language version in FindComputeCpp.cmake any more, it was occasionally conflicting with what downstream users were using. Instead it is set once globally in the root CMakeLists.txt file. Arguably there should be a better solution than this - should I try to grab the language version of the original source file, then set the integration header to that version?

cmake/Modules/FindComputeCpp.cmake
DuncanMcBain: I believe so. This functionality was added by Máté, he left that comment saying that the VS generator works differently. I've never encountered it personally but my CMake + Windows experience is rather more limited

cmake/Modules/FindComputeCpp.cmake
DuncanMcBain: It was after you explained to me again what the problem was I realised I should write it down in a way that made sense to me, in case I forget again!

cmake/Modules/FindComputeCpp.cmake
Ruyk: I am concern that Cmake will not set the right language version in this case since it is technically not a C++ file, although you are forcing it to be with the language property.  I tend not to trust CMake implicit decisions in most cases and mark it explicit, maybe by retrieving the property and setting it?

cmake/Modules/FindComputeCpp.cmake
Ruyk: It is a rather interesting situation, since in theory all properties are lists. I wonder if the underlying reason for using strings on the visual studio generator is because it was removing duplicated flags? If not, I think it should be a bug in CMake, since properties should be agnostics to targets

cmake/Modules/FindComputeCpp.cmake
DuncanMcBain: OK, I will add this.

cmake/Modules/FindComputeCpp.cmake
DuncanMcBain: It might be a CMake bug! We could test both ninja and VS with both versions to see if there's some common behaviour so we can settle on one or the other, but I'm not set up to do that here. This might be best placed in a different PR - could you make an issue for this?

mcleary:
I am actively testing this on Windows and it is working. The SYCL file gets compiled with the CPP being forced included into it. However, may I suggest the following:

```CMake
  # Force inclusion of the integration header for the host compiler
  if(MSVC)
    # Group SYCL files inside Visual Studio
    source_group("SYCL" FILES ${outputSyclFile})

    # Allow the source file to be edited using Visual Studio.
    # It will be added as a header file so it won't be compiled.
    set_property(SOURCE ${sourceFile} PROPERTY HEADER_FILE_ONLY true)
    target_sources(${targetName} PUBLIC ${sourceFile})

    # NOTE: The Visual Studio generators parse compile flags differently,
    # hence the different argument syntax
    if(CMAKE_GENERATOR MATCHES "Visual Studio")
      set(forceIncludeFlags "/FI\"${includedFile}\" /TP")
    else()
      set(forceIncludeFlags /FI ${includedFile} /TP)
    endif()    
  else()
      set(forceIncludeFlags "-include ${includedFile} -x c++")
  endif()
```
This allows the file to be edited inside Visual Studio, which is quite handy. It also groups SYCL files into a folder.

Ruyk:
Looks good to me @mcleary 
I guess the source_group can be outside of the if (MSVC) ? Other generators, such as the Eclipse one, may benefit from it...

cmake/Modules/FindComputeCpp.cmake
Ruyk: https://github.com/codeplaysoftware/computecpp-sdk/issues/79  done!

Ruyk:
According to https://cmake.org/cmake/help/v3.8/command/source_group.html  , it only works for visual studio, so ignore my comments ...

cmake/Modules/FindComputeCpp.cmake
DuncanMcBain: After discussion, it appears that because the language version is a property of the target, and because this code is similarly modifying the target, this should suffice (and indeed the flag -std=c++14 turns up in the output from CMake).

mcleary:
Please see #83 for some more changes. It contains all the changes in this branch plus a few more things.

DuncanMcBain:
Thanks all, and thank you @mcleary, this is a very good change now! And happy holidays!

Issue #80, opened 2017-12-20
Unable to build samples


domantasjurkus:
Hello, I am unable to build the samples provided with the repository.

![error](https://user-images.githubusercontent.com/9018827/34207003-5b8fb8ba-e591-11e7-84c1-c4e18e521132.png)
![cmake_version](https://user-images.githubusercontent.com/9018827/34206855-e0e51376-e590-11e7-84c6-c04a5da15298.png)
![computecpp_info](https://user-images.githubusercontent.com/9018827/34206861-e3b14188-e590-11e7-9f96-c9bbd44f178c.png)

Any advice is appreciated.

DuncanMcBain:
Hi @domantasjurkus,
instead of running
```sh
cmake ../samples
```
could you run
```sh
cmake ../
```
We'll update the documentation. Thanks!

domantasjurkus:
Thanks, that solved the issue with generating the Makefile.

However, a new error popped up when running `make`
```
dom@dom:~/computecpp-sdk/build$ make
[  1%] Built target simple-example-of-vectors_simple-example-of-vectors.cpp_0_ih
[  4%] Built target simple-example-of-vectors
[  5%] Built target images_images.cpp_0_ih
[  9%] Built target images
[ 10%] Built target custom-device-selector_custom-device-selector.cpp_0_ih
[ 14%] Built target custom-device-selector
[ 15%] Built target hello-world_hello-world.cpp_0_ih
[ 18%] Built target hello-world
[ 20%] Built target simple-vector-add_simple-vector-add.cpp_0_ih
[ 23%] Built target simple-vector-add
[ 24%] Built target example-sycl-application_example-sycl-application.cpp_0_ih
[ 28%] Built target example-sycl-application
[ 29%] Built target simple-private-memory_simple-private-memory.cpp_0_ih
[ 32%] Built target simple-private-memory
[ 34%] Built target reduction_reduction.cpp_0_ih
[ 37%] Built target reduction
[ 38%] Built target using-function-objects_using-function-objects.cpp_0_ih
[ 43%] Built target using-function-objects
[ 44%] Built target matrix-multiply_matrix-multiply.cpp_0_ih
[ 48%] Built target matrix-multiply
[ 49%] Built target simple-local-barrier_simple-local-barrier.cpp_0_ih
[ 52%] Built target simple-local-barrier
[ 54%] Built target parallel-for_parallel-for.cpp_0_ih
[...]

PR #81, opened 2017-12-20
Small fix: prepend std:: to exp

domantasjurkus:
Issue #80

### DIFF
--- a/samples/gaussian-blur/gaussian-blur.cpp
+++ b/samples/gaussian-blur/gaussian-blur.cpp
@@ -147,7 +147,7 @@ int main(int argc, char* argv[]) {
                                     itemID.get_global(1) + y);
             newPixel +=
                 inPtr.read(inputCoords, smpl) *
-                exp(-1.f * (x * x + y * y) / (2 * stddev * stddev)) /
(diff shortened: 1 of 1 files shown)


samples/gaussian-blur/gaussian-blur.cpp
```diff
@@ -147,7 +147,7 @@ int main(int argc, char* argv[]) {
                                     itemID.get_global(1) + y);
             newPixel +=
                 inPtr.read(inputCoords, smpl) *
-                exp(-1.f * (x * x + y * y) / (2 * stddev * stddev)) /
+                std::exp(-1.f * (x * x + y * y) / (2 * stddev * stddev)) /
```
DuncanMcBain: Could you change this to cl::sycl::exp? I have a funny feeling this might not work everywhere (for example, Windows, though I've not tested). Technically, there's no guarantee that std:: code will work inside kernels, though a large amount certainly can, so for this sample I'd prefer that it use the standard SYCL operation.

domantasjurkus: Sure, updated. Rebuilt to be sure, should be working.

PR #85, opened 2018-01-15
Update SDK to ComputeCpp v0.5.1

DuncanMcBain:
More 1.2.1 changes have been integrated, though we're not 100% there
yet. Old code will continue to work but will show deprecation messages
(which I don't think are useful to see when building the SDK).

### DIFF
--- a/.travis/build_computecpp.sh
+++ b/.travis/build_computecpp.sh
@@ -5,7 +5,7 @@ set -ev
 ###########################
 # Get ComputeCpp
 ###########################
-wget --no-check-certificate https://computecpp.codeplay.com/downloads/computecpp-ce/0.5.1/ubuntu-16.04-64bit.tar.gz
+wget --no-check-certificate https://computecpp.codeplay.com/downloads/computecpp-ce/0.5.1/ubuntu-14.04-64bit.tar.gz
 tar -xzf ubuntu-14.04-64bit.tar.gz -C /tmp
 mv /tmp/ComputeCpp-CE-0.5.1-Ubuntu-14.04-64bit /tmp/computecpp
 # Workaround for C99 definition conflict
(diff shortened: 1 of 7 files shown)


.travis/build_computecpp.sh
```diff
@@ -5,7 +5,8 @@ set -ev
 ###########################
 # Get ComputeCpp
 ###########################
-wget --no-check-certificate https://computecpp.codeplay.com/downloads/computecpp-ce/0.5.0/ubuntu-14.04-64bit.tar.gz
+wget --no-check-certificate https://computecpp.codeplay.com/downloads/computecpp-ce/0.5.1/ubuntu-14.04-64bit.tar.gz
 tar -xzf ubuntu-14.04-64bit.tar.gz -C /tmp
+mv /tmp/ComputeCpp-CE-0.5.1-Ubuntu-14.04-64bit /tmp/computecpp
```
ProGTX: What if the folder already exists?

Ruyk: why don't you import the new Khronos SYCL Parallel STL travis script? It also test with Intel OpenCL when available

samples/ivka/ivka.cpp
```diff
@@ -39,9 +39,11 @@ struct Bar : Foo {
 };
 
 using namespace cl::sycl;
+using m = cl::sycl::access::mode;
+using t = cl::sycl::access::target;
```
Ruyk: please don't use one letter shortcuts, at least use mode and target

.travis/build_computecpp.sh
DuncanMcBain: It won't, the machine is fresh every time.

DuncanMcBain: OK, I will add this in a separate CL I think. Thanks for the suggestion.

PR #92, opened 2018-02-22
Prefix any relative path to make it absolute

### LINKED ISSUE #87: FindComputeCpp.cmake cannot deal with relative paths
Currently, if you give a relative path to add_sycl_to_target, the device compiler won't be able to find the file (for example, you might be a directory build/, so all your sources are located one directory up). CMake handles this for the host side, and adjusts the path to be relative to the build directory, but FindComputeCpp.cmake does no such thing, so the device compiler can't find the file. We should fix this (or perhaps at the least warn when a relative path is passed to add_sycl_to_target).

Identified by #86.

jwlawson:
If `add_sycl_to_target` is called with relative paths to the source files
then cmake will not automatically adjust these to be absolute paths, and
therefore the device compiler is unable to find the source files. To get
around this, add a step to prefix the relative paths with the currrent
directory and so convert the relative path into an absolute path, which
will be found by the compiler.

Fixes #87


DuncanMcBain:
I didn't know about IS_ABSOLUTE. Thanks for that!

PR #96, opened 2018-03-02
Add step to remove any duplicate ComputeCpp flags

jwlawson:
When adding include directories it is easy to have the same include
directory linked to multiple cmake targets. Filter out any duplicate
include directories (and any other duplicate flags) before passing to
the compiler to keep the commands as terse as possible.


DuncanMcBain:
Thank you very much!

PR #103, opened 2018-04-15
Expose OpenMP support in samples in CMake

mmha:


mmha:
@DuncanMcBain tested this on his machine running 3.5 and 3.11. I also tested this code on 3.11.

[FindOpenMP](https://cmake.org/cmake/help/v3.0/module/FindOpenMP.html) is part of 3.0, so it should work on 3.2.

PR #111, opened 2018-05-11
WIP: Add option COMPUTECPP_USE_COMPILER_DRIVER.

mcleary:
- Adding the option COMPUTECPP_USE_COMPILER_DRIVER to enable/disable the usage of the ComputeCpp Compiler Driver.

- Add the usage of cmake_parse_arguments to the add_sycl_to_target function to allow more readable argument specification by named arguments.


mcleary:
I don't know why this fails on Travis. It is working just fine on both my work and home machines.

mmha:
You forgot to update the calls to `add_sycl_to_target` in the CMakeLists in the `test` directory. It probably works on your machine because you didn't set `COMPUTECPP_SDK_BUILD_TESTS` to `ON`.

mcleary:
@mmha good catch. Thanks. Should be fixed now.

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -251,7 +278,6 @@ function(__build_spir targetName sourceFile binaryDir fileCounter)
         ${device_compiler_includes})
     endforeach()
   endif()
-  list(REMOVE_DUPLICATES device_compiler_includes)
```
jwlawson: Is there any reason why you are removing this?

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -370,23 +414,48 @@ endfunction()
...
-  set(sourceFiles ${sourceFiles} ${ARGN})
   set(fileCounter 0)
   target_include_directories(
     ${targetName} SYSTEM
     PRIVATE ${OpenCL_INCLUDE_DIR}
     PRIVATE ${COMPUTECPP_INCLUDE_DIRECTORY}
   )
-  # Add custom target to run compute++ and generate the integration header
-  foreach(sourceFile ${sourceFiles})
-    if(NOT IS_ABSOLUTE ${sourceFile})
```
jwlawson: I think that you have removed this check to ensure that the source files have absolute paths. The custom compute++ command requires that all source files have absolute not relative paths.

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -370,23 +414,48 @@ endfunction()
...
+    # Set all flags required to compile the target when using the compiler driver
+    set(computeCppCompileOptions ${COMPUTECPP_DEVICE_COMPILER_FLAGS})
+    get_property(includeAfter TARGET ${targetName} PROPERTY COMPUTECPP_INCLUDE_AFTER)
+    if(includeAfter)
+      list(APPEND computeCppCompileOptions -fsycl-ih-last)
     endif()
-    __build_spir(${targetName} ${sourceFile} ${binaryDir} ${fileCounter})
-    MATH(EXPR fileCounter "${fileCounter} + 1")
-  endforeach()
+    target_compile_options(${targetName} PUBLIC ${computeCppCompileOptions})
```
jwlawson: This does not add the `COMPUTECPP_USER_FLAGS`. Is this intentional?

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -370,23 +414,48 @@ endfunction()
...
+  else()
+    if(NOT binaryDir)
+      message(WARNING " ComputeCpp Warning: BINARY_DIR was not specified. Using CMAKE_CUURENT_BINARY_DIR which is \"${CMAKE_CUURENT_BINARY_DIR}\"")
+      set(binaryDir ${CMAKE_CURRENT_BINARY_DIR})
+    endif()
+    if(NOT sourceFiles)
+      message(FALTAL_ERROR " ComputeCpp Error: SOURCES was not specified. At least one source file must be provided.")
```
jwlawson: Is this really an error? I can think of a use case where a user wants to link against ComputeCPP but doesn't actually need to compile any integration headers. One way of doing that would be to call `add_sycl_to_target` without specifying any source files (and so also not needing to specify a binary dir).

You would need to add a check before the `foreach` below though.

If you do decide to keep this error, fix the typo `FALTAL_ERROR` -> `FATAL_ERROR`.

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -370,23 +414,48 @@ endfunction()
...
+  set(oneValueArgs TARGET BINARY_DIR)
+  set(multiValueArgs SOURCES)
+  cmake_parse_arguments(add_sycl_to_target "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
+
+  set(targetName ${add_sycl_to_target_TARGET})
+  set(binaryDir ${add_sycl_to_target_BINARY_DIR})
+  set(sourceFiles ${add_sycl_to_target_SOURCES})
 
-  set(sourceFiles ${sourceFiles} ${ARGN})
   set(fileCounter 0)
```
jwlawson: The filecounter is only used when not using the driver. This could be moved down into the `else` branch below.

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -370,23 +414,48 @@ endfunction()
 #  binaryDir : Intermediate directory to output the integration header.
 #  sourceFiles : Source files to be compiled for SYCL.
 #
-function(add_sycl_to_target targetName binaryDir sourceFiles)
+function(add_sycl_to_target)
+  set(options)
+  set(oneValueArgs TARGET BINARY_DIR)
+  set(multiValueArgs SOURCES)
+  cmake_parse_arguments(add_sycl_to_target "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
```
jwlawson: To be extra cautious, you could add a check after this to make sure there are no unexpected arguments. (Check for `<prefix>_UNPARSED_ARGUMENTS`)

CMakeLists.txt
```diff
@@ -16,6 +16,10 @@ include_directories(SYSTEM ${COMPUTECPP_INCLUDE_DIRECTORY})
 
 enable_testing()
 
+# This will enable ComputeCpp compiler driver if COMPUTECPP_USE_COMPILER_DRIVER
+# was specified and the driver is supported by the current system.
+enable_sycl_compiler_driver()
```
jwlawson: I might be missing something here, but I don't think that having to explicitly enable the compiler driver is a particularly good idea, given that it can be enabled/disabled by the user through the cmake cache. Why should someone developing SYCL applications care whether the driver is used or not?

I can just about see a use case, if a developer wishes to have some executables/libraries compiled with the host compiler, then others compiled with the driver, but that would surely require a `disable_sycl_compiler_driver` method too?

I think that this would be better inlined at [when you specify handle the use driver option](https://github.com/codeplaysoftware/computecpp-sdk/pull/111/files#diff-3b92b8ec68465519e2a6b06149381084R223) so that the compiler driver is enabled when the developer calls `find_package(computecpp)`.

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -370,23 +414,48 @@ endfunction()
 #  binaryDir : Intermediate directory to output the integration header.
 #  sourceFiles : Source files to be compiled for SYCL.
```
jwlawson: Can you change the documentation to match the new function signature?

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -212,6 +218,27 @@ define_property(
   non-standards-conformant SYCL code to compile with ComputeCpp."
 )
 
+# Enable the usage of the ComputeCpp compiler driver. This removes the need
+# to explicitly generate the sycl integration header in a separate compiler pass.
+if(COMPUTECPP_USE_COMPILER_DRIVER)
+  # Check if the current system supports the compiler driver
+  set(isCompilerDriverSupported NO)
+
+  # For now, the compiler driver is only supported if the host compiler is either clang or gcc
```
jwlawson: Is this to ensure that any additional compiler flags are recognised? You are going to override the `CXX` compiler, so I'm not sure I understand this requirement?

mcleary:
@jwlawson Thanks for the comments. I will address them as soon as I can.

mcleary:
I still have some changes to do in this pull request. I just found out that the flags `-emit-llvm` and `-sycl` are not compatible with `-sycl-driver`, since the driver will manage the use of these flags internally. I need change the CMake code to take this into account.

mcleary:
I am going to close this pull request for now because @mmha has a much nicer solution for this.

PR #113, opened 2018-05-15
Fix const capture crash on Windows CMake builds

### LINKED ISSUE #112: -fno-ms-compatibility causes hard to find bugs with lambda const capture
I just spent a considerable amount of time trying to figure out why this kernel crashes somewhere inside ComputeCpp's `process_functor_arguments`:

```c++
my_queue.submit([&](cl::sycl::handler& cgh) {
  auto out = my_buf.get_access<cl::sycl::access::mode::write>(cgh);
  const int some_const_value = 16;
  auto my_kernel = [=](cl::sycl::item<1> item) {
    int cause_capture = some_const_value;
    out[item] = 0.f;
  };
  std::cout << "sizeof(my_kernel) = " << sizeof(my_kernel) << std::endl;
  cgh.parallel_for<class const_crash>(cl::sycl::range<1>(64), my_kernel);
});
```

As it turns out, this only happens on Windows, and only when using a CMake build (not the Visual Studio project wizard).

The culprit becomes apparent when comparing the `*.sycl` files generated by the device compiler.

CMake on Windows:
```
const size_t kernel_info< const_crash >::arg_size = 3;
const size_t kernel_info< const_crash >::arg_sizes[] = {
  8, // __attribute__((address_space(1))) float *
  16, // class cl::sycl::device_index_array
  16, // class cl::sycl::device_index_array
  0
};
```

Visual Studio wizard:
```
const size_t kernel_info< const_crash >::arg_size = 4;
const size_t kernel_info< const_crash >::arg_sizes[] = {
  4, // const int
  8, // __attribute__((address_space(1))) float *
  16, // class cl::sycl::device_index_array
  16, // class cl::sycl::device_index_array
  0
};
```

It appears that clang propagates the `const` value into the lambda, while MSVC (presumably) does not. This causes a mismatch between the lambda and the kernel at runtime.

When comparing the device compiler flags used by the two projects, I found the `-fno-ms-compatibility` flag set for CMake builds [here](https://github.com/codeplaysoftware/computecpp-sdk/blob/a8808b4adbe28e344f1637d6fb18d88df4d57b04/cmake/Modules/FindComputeCpp.cmake#L69). And, lo and behold, removing this flag fixes the issue.

I'm not sure if the original reasons for including this flag are still relevant, nor the specifics of the issue, but I tried to compile the program above with `const char16_t some_const_value = 16;`, which appeared to work just fine.

Tested with ComputeCpp 0.6.1, 0.7.0 and 0.8.0, on Windows 10 and Ubuntu 16.04.

psalz:
This fixes #112.

All but one sample are confirmed to still work on Windows after the change. The one that's not working is `vptr`, however, the same crash already occurred before the change. One additional caveat: Since I don't have a compatible GPU available on my Windows machine, I had to change the `gpu_selector` in `opencl-c-interop` to a `cpu_selector`.

This is also a necessary (but not sufficient) change in order to make the tests build on Windows again. I'll submit another PR for this in a minute.

Lastly, note that I had to roll back from ComputeCpp 0.8.0 to 0.7.0, since the former seems to have broken some of the samples (also on Linux).

### DIFF
--- a/cmake/Modules/FindComputeCpp.cmake
+++ b/cmake/Modules/FindComputeCpp.cmake
@@ -62,16 +62,6 @@ mark_as_advanced(COMPUTECPP_DISABLE_GCC_DUAL_ABI)
 set(COMPUTECPP_USER_FLAGS "" CACHE STRING "User flags for compute++")
 mark_as_advanced(COMPUTECPP_USER_FLAGS)
 
-# Platform-specific arguments
-if(MSVC)
-  # If you encounter problems with types such as char16_t and char32_t not
-  # being defined, you can try to enable the flag below.
-  # See also:
-  # https://github.com/codeplaysoftware/computecpp-sdk/pull/51#discussion_r139399093
-  # https://github.com/codeplaysoftware/computecpp-sdk/issues/112
-  # set (COMPUTECPP_PLATFORM_SPECIFIC_ARGS "-fno-ms-compatibility")
-endif()
-
 # Find OpenCL package
 find_package(OpenCL REQUIRED)
 
@@ -284,7 +274,6 @@ function(__build_spir targetName sourceFile binaryDir fileCounter)
     COMMAND ${COMPUTECPP_DEVICE_COMPILER}
             ${COMPUTECPP_DEVICE_COMPILER_FLAGS}
             -isystem ${COMPUTECPP_INCLUDE_DIRECTORY}
-            ${COMPUTECPP_PLATFORM_SPECIFIC_ARGS}
             ${device_compiler_includes}
             -o ${outputSyclFile}
             -c ${sourceFile}
--- a/cmake/Modules/FindComputeCpp.cmake
+++ b/cmake/Modules/FindComputeCpp.cmake
@@ -64,9 +64,12 @@ mark_as_advanced(COMPUTECPP_USER_FLAGS)
 
 # Platform-specific arguments
 if(MSVC)
-  # Workaround to an unfixed Clang bug, rationale:
+  # If you encounter problems with types such as char16_t and char32_t not
+  # being defined, you can try to enable the flag below.
+  # See also:
   # https://github.com/codeplaysoftware/computecpp-sdk/pull/51#discussion_r139399093
-  set (COMPUTECPP_PLATFORM_SPECIFIC_ARGS "-fno-ms-compatibility")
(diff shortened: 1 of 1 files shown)


DuncanMcBain:
That's very fast - I was about to push some of the same fixes, thanks! I'll take a look at this as it was only when I grabbed a colleague with a Visual Studio install that I realised there were some... not so great bits of CMake causing problems!

Honestly, my preferred change at this stage is to remove the thing entirely. If users run into issues they can contact us and we can provide a fix - but I've not been able to find any cases where this breaks code (whereas we've now had two cases of code breaking with -fno-ms-compatibility).

I'm aware of the slight breakages with 0.8.0 - since you had flagged this up, I wanted to get it in before I push and tag the next release.

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -64,9 +64,12 @@ mark_as_advanced(COMPUTECPP_USER_FLAGS)
...
 # Platform-specific arguments
 if(MSVC)
-  # Workaround to an unfixed Clang bug, rationale:
+  # If you encounter problems with types such as char16_t and char32_t not
+  # being defined, you can try to enable the flag below.
+  # See also:
   # https://github.com/codeplaysoftware/computecpp-sdk/pull/51#discussion_r139399093
-  set (COMPUTECPP_PLATFORM_SPECIFIC_ARGS "-fno-ms-compatibility")
+  # https://github.com/codeplaysoftware/computecpp-sdk/issues/112
+  # set (COMPUTECPP_PLATFORM_SPECIFIC_ARGS "-fno-ms-compatibility")
```
DuncanMcBain: To clarify, I would remove this entire if/else, and also remove the reference to this variable on line 284, since this is the only place it is used.

psalz:
Should I then remove the `COMPUTECPP_PLATFORM_SPECIFIC_ARGS` variable entirely, as it isn't being used anywhere else?

Also as a heads up: I went back and tried running the `vptr` sample again as a release build, and it works (so it's likely the same issue as with the `vptr/offset.cc` tests).

cmake/Modules/FindComputeCpp.cmake
psalz: Nevermind my comment then, will do!

DuncanMcBain:
Yup that's great, thanks!

PR #114, opened 2018-05-15
Fix Windows tests

psalz:
This makes the tests compile (and pass) on Windows again.

The changes made:
* Remove the `-fno-ms-compatibility` flag. Without this, Clang on Windows will report the following errors:
    ```
    error G30564CF7: use of undeclared identifier 'chdir'; did you mean '_chdir'?
    error G30564CF7: use of undeclared identifier 'fdopen'
    error G30564CF7: use of undeclared identifier 'read'
    error G30564CF7: use of undeclared identifier 'write'
    error G30564CF7: use of undeclared identifier 'close'
    ```
    See #113 for more information on this change.
* Remove hardcoded Google Test library paths (such as `${gtest_BINARY_DIR}/libgtest.a`), and use CMake's cross-platform "Threads" instead of using `pthread` directly.
* Finally, the C++ language version for the `legacy-pointer` tests is set from 11 to 14 (the project default). I'm not sure why this was set in the first place, as the tests seem to run just fine without it. However,  when compiling as C++11 on Windows, Clang will complain about missing trailing return types in some MS headers (e.g. `Microsoft Visual Studio 14.0\VC\include\xstddef(732,2): error G90440305: 'auto' return without trailing return type; deduced return types are a C++14 extension`).

All tests are green on Windows, except for `vptr/offset.cc` in debug builds, which [is a known issue](https://github.com/codeplaysoftware/computecpp-sdk/issues/112#issuecomment-388954438).

Note that I couldn't get the experimental `vptr/runtime.cc`  tests (enabled with `-DCOMPUTECPP_INTERFACE=1`) to run, so I didn't touch those.

All tests are also confirmed to still work on Linux (on Ubuntu 16.04, with a Tesla K20m via [pocl](https://github.com/pocl/pocl)).


psalz:
Well, it appears like the `legacy_basic` test is failing with Clang builds on Travis for some reason. While I did only test it with GCC (7), I just went back and tried Clang-5.0. Works for me!

DuncanMcBain:
OK thanks, I'll take a look. Travis flakes out sometimes...

DuncanMcBain:
OK I reran it and it works. That either means we have a race condition somewhere or Travis is a little sketchy (and at the moment, I'm inclined to believe it is the latter, as local testing has never revealed any problems with this particular code).

tests/vptr/CMakeLists.txt
```diff
@@ -25,9 +25,9 @@ if (COMPUTECPP_INTERFACE)
   add_definitions(-DCOMPUTECPP_INTERFACE=${COMPUTECPP_INTERFACE})
   add_executable(runtime runtime.cc)
   set_property(TARGET runtime PROPERTY CXX_STANDARD 11)
```
DuncanMcBain: Would you be able to remove this as well? It's not needed, setting the language level to 14 in the base CMakeLists.txt should be enough :)

psalz: Done!

DuncanMcBain:
These fixes are great, thank you so much!

PR #118, opened 2018-06-14
Modernise FindComputeCpp.cmake

DuncanMcBain:
This is a very large update to FindComputeCpp.cmake that brings with
it (regrettably) some API changes. See the samples for details, but
the short story is that the add_sycl_to_target function now takes
named parameters, which has helped a lot. Additionally, there have
been bugfixes in the internals of the module, meaning that the include-
after property is more well-behaved.

The other change that is most likely to affect users is the way in which
different bitcode outputs are selected. Previously different options
like "USE_PTX" or similar could be specified, but now an actual value
must be set, and defaults to spir64. This requires a little more work
on the part of the user when a different kind of IR is needed but is
more robust.


rodburns:
How do I specify output for PTX or SPIR-V now? Or is it just spir and spir64?

DuncanMcBain:
There's a flag, COMPUTECPP_BITCODE, which you set to any of spir[64], spirv[64] or ptx[64] (i.e. the 64 is optional, with spir64 being the default). I should add this to the readme, I think - will do so.

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -204,51 +216,76 @@ define_property(
...
+    ${ARGN}
+  )
+  get_filename_component(sourceFileName ${SDK_BUILD_IR_SOURCE} NAME)
 
-  # Set the path to the Sycl file.
-  set(outputSyclFile ${binaryDir}/${sourceFileName}.sycl)
+  # Set the path to the integration header.
+  set(outputSyclFile ${SDK_BUILD_IR_BINARY_DIR}/${sourceFileName}.sycl)
 
   # Add any user-defined include to the device compiler
```
mmha: Unfortunately this won't work for every build script:

```cmake
add_executable(app main.cpp)
add_sycl_to_target(TARGET app BINARY_DIR ${CMAKE_CURRENT_BINARY_DIR} SOURCES main.cpp)
target_include_directories(app PRIVATE ${SOMETHING_INCLUDE_DIRS}) # won't find this
target_link_libraries(app PRIVATE Boost::boost) # or boost
```

The include directories must be gathered lazily with a generator expression to make `add_sycl_to_target()` work with property propagation through `target_link_libraries()`. The same is true for compile definitions.

Compile options are a special case for us because we are invoking a different compiler so we should not copy all the compiler flags automatically in my opinion (`-g` or `-mavx` come to mind).

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -276,39 +370,49 @@ function(__build_spir targetName sourceFile binaryDir fileCounter)
             -isystem ${COMPUTECPP_INCLUDE_DIRECTORY}
             ${device_compiler_includes}
             -o ${outputSyclFile}
-            -c ${sourceFile}
-    DEPENDS ${sourceFile}
-    IMPLICIT_DEPENDS CXX ${sourceFile}
-    WORKING_DIRECTORY ${binaryDir}
+            -c ${SDK_BUILD_IR_SOURCE}
+    DEPENDS ${ir_dependencies}
+    IMPLICIT_DEPENDS CXX ${SDK_BUILD_IR_SOURCE}
```
mmha: What is `CXX` here?

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -276,39 +370,49 @@ function(__build_spir targetName sourceFile binaryDir fileCounter)
...
+    string(REPLACE "${CMAKE_CURRENT_SOURCE_DIR}/" ""
+      rel_source_file ${SDK_BUILD_IR_SOURCE}
+    )
+    list(REMOVE_ITEM current_sources ${rel_source_file})
+    # Add SYCL header to source list
+    list(APPEND current_sources ${outputSyclFile})
+    set_property(TARGET ${SDK_BUILD_IR_TARGET}
+      PROPERTY SOURCES ${current_sources})
     # CMake/gcc don't know what language a .sycl file is, so tell them
     set_property(SOURCE ${outputSyclFile} PROPERTY LANGUAGE CXX)
```
mmha: It should also have the [HEADER_FILE_ONLY](https://cmake.org/cmake/help/v3.12/prop_sf/HEADER_FILE_ONLY.html) property.

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -358,31 +454,40 @@ endfunction()
 #  Adds a SYCL compilation custom command associated with an existing
 #  target and sets a dependancy on that new command.
 #
-#  targetName : Name of the target to add a SYCL to.
-#  binaryDir : Intermediate directory to output the integration header.
-#  sourceFiles : Source files to be compiled for SYCL.
+#  TARGET : Name of the target to add SYCL to.
+#  BINARY_DIR : Intermediate directory to output the integration header.
```
mmha: Is this argument necessary? Can't we just use `CMAKE_CURRENT_BINARY_DIR` instead?

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -358,31 +454,40 @@ endfunction()
...
     MATH(EXPR fileCounter "${fileCounter} + 1")
   endforeach()
-
-  # Link with the ComputeCpp runtime library
-  target_link_libraries(${targetName} PUBLIC $<$<OR:$<CONFIG:Debug>,$<CONFIG:RelWithDebInfo>>:${COMPUTECPP_RUNTIME_LIBRARY_DEBUG}>
-                                             $<$<NOT:$<OR:$<CONFIG:Debug>,$<CONFIG:RelWithDebInfo>>>:${COMPUTECPP_RUNTIME_LIBRARY}>
-                                             ${OpenCL_LIBRARIES})
-
+  target_link_libraries(${SDK_ADD_SYCL_TARGET}
+    PUBLIC ComputeCpp::ComputeCpp
```
mmha: This may be a feature for a future PR, but (optionally?) appending ComputeCpp to the target's `BUILD_RPATH` might be a useful addition so users won't have to set `LD_LIBRARY_PATH` to test their applications.

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -51,18 +50,13 @@ elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
...
-option(COMPUTECPP_DISABLE_GCC_DUAL_ABI "Compile with pre-5.1 ABI" OFF)
-mark_as_advanced(COMPUTECPP_DISABLE_GCC_DUAL_ABI)
-
 set(COMPUTECPP_USER_FLAGS "" CACHE STRING "User flags for compute++")
 mark_as_advanced(COMPUTECPP_USER_FLAGS)
 
-# Find OpenCL package
+set(COMPUTECPP_BITCODE "spir64" CACHE STRING
+  "Bitcode type to use as SYCL target in compute++")
+mark_as_advanced(COMPUTECPP_BITCODE)
```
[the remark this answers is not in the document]
DuncanMcBain: I understand that this could be very useful, but I can't think of that many situations where you'd want to build different targets in the same project with different bitcode formats (then there are questions like what happens when those targets are part of one object, but maybe it would just work?).

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -154,21 +149,26 @@ else()
...
+  OUTPUT_VARIABLE COMPUTECPP_PLATFORM_IS_SUPPORTED
+  RESULT_VARIABLE COMPUTECPP_INFO_TOOL_RESULT OUTPUT_STRIP_TRAILING_WHITESPACE)
+if(NOT COMPUTECPP_INFO_TOOL_RESULT EQUAL "0")
+  message(FATAL_ERROR "platform - Error checking platform support!")
+else()
+  mark_as_advanced(COMPUTECPP_PLATFORM_IS_SUPPORTED)
+  if (COMPUTECPP_PLATFORM_IS_SUPPORTED)
+    message(STATUS "platform - your system can support ComputeCpp")
+  else()
+    message(STATUS "platform - your system CANNOT support ComputeCpp")
```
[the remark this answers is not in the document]
DuncanMcBain: Done!

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -177,21 +177,33 @@ else()
...
+    list(APPEND COMPUTECPP_DEVICE_COMPILER_FLAGS --gcc-toolchain=${SDK_TOOLCHAIN_DIR})
   endif()
+  list(APPEND COMPUTECPP_DEVICE_COMPILER_FLAGS --sysroot=${SDK_SYSROOT_DIR})
+  list(APPEND COMPUTECPP_DEVICE_COMPILER_FLAGS -target ${SDK_TARGET_TRIPLE})
+endif()
+
+separate_arguments(COMPUTECPP_DEVICE_COMPILER_FLAGS)
+
+if(NOT TARGET OpenCL::OpenCL)
+  add_library(OpenCL::OpenCL IMPORTED UNKNOWN)
```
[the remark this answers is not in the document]
DuncanMcBain: Yes, that's all we use. I'll change it anyway, since you are correct.

cmake/Modules/FindComputeCpp.cmake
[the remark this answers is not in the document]
mmha: It's more of a consistency thing with the rest of CMake. `COMPUTECPP_BITCODE` would set the default value of a target property with the same name, just like `CMAKE_CXX_STANDARD` sets the default value for the `CXX_STANDARD` target property.

I understand that this can be seen as a new feature and adding a target property shouldn't break backwards compatibility, so we can postpone this to a later PR if you like.

> then there are questions like what happens when those targets are part of one object, but maybe it would just work?

This can be handled by adding a potential `COMPUTECPP_BITCODE` target property to [`COMPATIBLE_INTERFACE_STRING`](https://cmake.org/cmake/help/v3.12/prop_tgt/COMPATIBLE_INTERFACE_STRING.html) (see also [here](https://cmake.org/cmake/help/v3.12/manual/cmake-buildsystem.7.html#compatible-interface-properties)).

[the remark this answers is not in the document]
DuncanMcBain: OK, on reflection, I think it's best that we make an issue for this, then add it as a later feature (:

cmake/Modules/FindComputeCpp.cmake
DuncanMcBain: https://cmake.org/cmake/help/v3.2/command/add_custom_command.html

It says "The language given specifies the programming language whose corresponding dependency scanner should be used." In this case, use the CXX rules.

cmake/Modules/FindComputeCpp.cmake
DuncanMcBain: Probably, assuming that most callers of "add_sycl_to_target" simply put CMAKE_CURRENT_BINARY_DIR in there anyway. I'll do this.

cmake/Modules/FindComputeCpp.cmake
DuncanMcBain: This happens for executables built on the machine, but if you install them they'll be stripped, as far as I can tell.

cmake/Modules/FindComputeCpp.cmake
DuncanMcBain: Done! But only in the "non-include-after" case.

DuncanMcBain:
There are now more and deeper changes. I will add a further update to the README to explain what is different.

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -204,51 +194,63 @@ define_property(
...
-    endforeach()
-  endif()
-  list(REMOVE_DUPLICATES device_compiler_includes)
+  set(include_directories "$<TARGET_PROPERTY:${SDK_BUILD_IR_TARGET},INCLUDE_DIRECTORIES>")
+  set(compile_definitions "$<TARGET_PROPERTY:${SDK_BUILD_IR_TARGET},COMPILE_DEFINITIONS>")
+  set(generated_include_directories
+    $<$<BOOL:${include_directories}>:-I\"$<JOIN:${include_directories},\"\t-I\">\">)
+  set(generated_compile_definitions
+    $<$<BOOL:${compile_definitions}>:-D$<JOIN:${compile_definitions},\t-D>>)
+  target_include_directories(${SDK_BUILD_IR_TARGET} PRIVATE ${OpenCL_INCLUDE_DIRS})
```
jwlawson: This should be handled later on in the call to `target_link_libraries(... ComputeCpp::ComputeCpp)`.

If you need to explicitly add the OpenCL include directories to the custom command, then I would do that in the same way we add the ComputeCpp include directories. Though possibly with the generator expressions this won't be needed any more?

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -261,54 +263,83 @@ function(__build_spir targetName sourceFile binaryDir fileCounter)
...
     ${COMPUTECPP_DEVICE_COMPILER_FLAGS}
-    ${COMPUTECPP_USER_FLAGS})
-  # Convert argument list format
+    ${COMPUTECPP_USER_FLAGS}
+    ${target_compile_flags}
+  )
   separate_arguments(COMPUTECPP_DEVICE_COMPILER_FLAGS)
 
+  set(ir_dependencies ${SDK_BUILD_IR_SOURCE})
+  if(target_libraries)
```
jwlawson: Somewhere in your changes you removed the code that would set this variable. This could mean that library dependencies aren't handled correctly in the custom command.

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -204,51 +194,63 @@ define_property(
...
-  endforeach()
-  if (CMAKE_INCLUDE_PATH)
-    foreach(directory ${CMAKE_INCLUDE_PATH})
-      set(device_compiler_includes "-I${directory}"
-        ${device_compiler_includes})
-    endforeach()
-  endif()
-  list(REMOVE_DUPLICATES device_compiler_includes)
+  set(include_directories "$<TARGET_PROPERTY:${SDK_BUILD_IR_TARGET},INCLUDE_DIRECTORIES>")
+  set(compile_definitions "$<TARGET_PROPERTY:${SDK_BUILD_IR_TARGET},COMPILE_DEFINITIONS>")
```
jwlawson: As well as include directories and compile definitions, it might be worth looking at copying compile options too.

cmake/Modules/FindComputeCpp.cmake
DuncanMcBain: You're right, I can just remove this. I was wondering why a certain directory was turning up twice, this explains a lot. Thanks!

cmake/Modules/FindComputeCpp.cmake
DuncanMcBain: We stopped short of adding the compile options because of things that might not play so nicely - like -mavx, or -pthread, or similar (we being @mmha and me). I could check whether these will be ignored by the device compiler or whether we should avoid adding them for now.

DuncanMcBain:
@jwlawson you are correct about the code styling issues. It needs a pass to clean this up. Similarly there's a cmake code formatter on Github that I've been tempted to download and use to get everything consistent, because at the moment it is anything but!

Once I have merged these changes, I will port them to the SYCL-DNN copy of FindComputeCpp, to try to further reduce the divergence. Thanks for taking a look at this :)

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -51,146 +51,136 @@ elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
...
   NAMES ComputeCpp ComputeCpp_vs2015_d
-  PATHS ${COMPUTECPP_PACKAGE_ROOT_DIR}
-  HINTS ${COMPUTECPP_PACKAGE_ROOT_DIR}/lib PATH_SUFFIXES lib
-  DOC "ComputeCpp Debug Runtime Library" NO_DEFAULT_PATH)
-
-if (EXISTS ${COMPUTECPP_RUNTIME_LIBRARY_DEBUG})
-  mark_as_advanced(COMPUTECPP_RUNTIME_LIBRARY_DEBUG)
-else()
-  message(FATAL_ERROR "ComputeCpp Debug Runtime Library - Not found!")
+  PATHS ${ComputeCpp_ROOT_DIR}
```
mmha: the `PATHS` in all `find_*()` calls are redundant.

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -51,146 +51,136 @@ elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
...
+  HINTS ${ComputeCpp_ROOT_DIR}/lib
+  PATH_SUFFIXES lib
+  DOC "ComputeCpp Debug Runtime Library")
+
+find_path(ComputeCpp_INCLUDE_DIRS
+  NAMES "CL/sycl.hpp"
+  HINTS ${ComputeCpp_ROOT_DIR}/include
+  DOC "The ComputeCpp include directory")
+get_filename_component(ComputeCpp_INCLUDE_DIRS ${ComputeCpp_INCLUDE_DIRS} ABSOLUTE)
```
mmha: I'm pretty sure the filename is absolute already.

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -51,146 +51,136 @@ elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
...
+
+execute_process(COMMAND ${ComputeCpp_INFO_EXECUTABLE} "--dump-version"
+  OUTPUT_VARIABLE ComputeCpp_VERSION
```
mmha: If we remove the edition prefix with a regex and follow semver users can write this:

```cmake
find_package(ComputeCpp 0.8 REQUIRED)
find_package(ComputeCpp 0.8 EXACT REQUIRED)
```

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -23,14 +23,14 @@
...
 #
 #  Latest version of this file can be found at:
 #    https://github.com/codeplaysoftware/computecpp-sdk
 
-# Require CMake version 3.2.2 or higher
 cmake_minimum_required(VERSION 3.2.2)
+include(FindPackageHandleStandardArgs)
 
 # Check that a supported host compiler can be found
 if(CMAKE_COMPILER_IS_GNUCXX)
```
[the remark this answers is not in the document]
DuncanMcBain: Yup - done.

cmake/Modules/FindComputeCpp.cmake
DuncanMcBain: Yeah I'll stick to hints. Done.

cmake/Modules/FindComputeCpp.cmake
DuncanMcBain: This will get relegated to a later issue, too.

cmake/Modules/FindComputeCpp.cmake
DuncanMcBain: If we can later prove that it is in all cases, it's an easy fix - 'til then, it doesn't hurt to have it in.

PR #119, opened 2018-06-19
Adding tiled convolution to SYCL.

mehdi-goli:
Creating tiled-based convolution.


samples/tiled-convolution/tiled-convolution.cpp
```diff
@@ -0,0 +1,307 @@
...
+      : fil_acc(fil_acc_),
+        in_acc(in_acc_),
+        out_acc(out_acc_),
+        total_size(total_size_),
+        fil_size(fil_size_) {}
+  void inline operator()(cl::sycl::nd_item<2> item_id) {
+    index_t id_m = item_id.get_global(0);
+    index_t id_n = item_id.get_global(1);
+    index_t m, f_m;
+    index_t n, f_n;
```
Ruyk: what is f_m, f_n ? can you use more elaborated names for the variables?

samples/tiled-convolution/tiled-convolution.cpp
```diff
@@ -0,0 +1,307 @@
...
+template <typename index_t>
+class matrix_size_t {
+ public:
+  const index_t m;
+  const index_t n;
+  matrix_size_t(const index_t m_, const index_t n_) : m(m_), n(n_) {}
+  inline index_t size() const { return (m * n); }
+};
+
+// the tiled based convolution functor
```
Ruyk: can you add a comment explaining how the convolution algorithm works? If you have a separate README somewhere  or an image then you can just add that to the pull request to help understand the algorithm

samples/tiled-convolution/tiled-convolution.cpp
```diff
@@ -0,0 +1,307 @@
...
+        total_size(total_size_),
+        fil_size(fil_size_) {}
+  void inline operator()(cl::sycl::nd_item<2> item_id) {
+    index_t id_m = item_id.get_global(0);
+    index_t id_n = item_id.get_global(1);
+    index_t m, f_m;
+    index_t n, f_n;
+    typename write_accessor_t::value_type val = 0.0;
+#ifdef __SYCL_DEVICE_ONLY__
+#pragma umroll
```
Ruyk: are you sure is "umroll" ? :-) i think its unroll. 
For which platform do you need this unroll? or is it beneficial across all range of CE platforms?

samples/tiled-convolution/tiled-convolution.cpp
```diff
@@ -0,0 +1,307 @@
...
+    }
+    out_acc[id_m][id_n] = val / fil_size.size();
+  }
+};
+
+// round up function to construct the global and local size
+template <typename index_t>
+inline index_t round_up(const index_t x, const index_t y) {
+  return ((((x) + (y) -1) / (y)) * (y));
+}
```
Ruyk: Why so many parenthesis? was this a macro before?

samples/tiled-convolution/tiled-convolution.cpp
```diff
@@ -0,0 +1,307 @@
...
+  // mask filter value
+  auto filter_data = data_t(0.3);
+  // input array
+  std::vector<data_t> input(total_buffer.size(), input_data);
+  // mask array
+  std::vector<data_t> filter(fil_size.size(), filter_data);
+
+  auto property_list =
+      cl::sycl::property_list{cl::sycl::property::queue::enable_profiling()};
+  // constructing a SYCL queue for CVengine OpenCL device where
```
Ruyk: This is not just for the CVengine as it is written, can you remove this comment

samples/tiled-convolution/tiled-convolution.cpp
```diff
@@ -0,0 +1,307 @@
...
+            << " , avarage_application_execution_time, "
+            << avarage_application_execution_time << "\n";
+#endif
+  // check the correctness of the result
+  auto out_data = out_buff.get_access<read_t>();
+  for (index_t m = 0; m < total_buffer.m; m++)
+    for (index_t n = 0; n < total_buffer.m; n++)
+      if (!(std::abs(index_t(out_data[m][n] - (input_data * filter_data))) <
+            1e-4))
+        std::cout << " The result is wrong " << std::endl;
```
Ruyk: please use braces for all for and ifs
Also, if the result is wrong, return a non 0 value from the main so ctest can catch the potential errors.

samples/tiled-convolution/tiled-convolution.cpp
```diff
@@ -0,0 +1,310 @@
...
+                             cl::sycl::id<2>(host_offset_m, host_offset_n)),
+                         conv<read_accessor_t, write_accessor_t, index_t,
+                              matrix_size_t<index_t>>(fil_acc, in_acc, out_acc,
+                                                      total_buffer, fil_size));
+      });
+// this part is used to profile each tiled kernel
+#ifdef PROFILE_SYCL
+      auto i = n + m * num_host_tile_n;
+      event.wait();
+      auto end = std::chrono::system_clock::now();
```
DuncanMcBain: The first iteration take longer than the others, because it will include the module compilation time. You could use the program class to avoid this (though some implementations might still have compiling to do, it should still help).

samples/tiled-convolution/tiled-convolution.cpp
```diff
@@ -0,0 +1,310 @@
...
+          cl::sycl::info::event_profiling::command_start>();
+      auto end_time = event.get_profiling_info<
+          cl::sycl::info::event_profiling::command_end>();
+      auto current_submission_time = (start_time - submit_time) / 1000000.0f;
+      auto current_execution_time = (end_time - start_time) / 1000000.0f;
+
+      total_execution_time += current_execution_time;
+      total_submission_time += current_submission_time;
+#ifdef PER_EVENT_PROFILING
+      std::cout << "event, " << i << " , current_kernel_submission_time, "
```
DuncanMcBain: I don't think we should have as much console output as this, it looks quite untidy.

samples/tiled-convolution/tiled-convolution.cpp
```diff
@@ -0,0 +1,310 @@
...
+// this part is used to profile each tiled kernel
+#ifdef PROFILE_SYCL
+      auto i = n + m * num_host_tile_n;
+      event.wait();
+      auto end = std::chrono::system_clock::now();
+      std::chrono::duration<double, std::milli>
+          current_application_execution_time = end - start;
+      total_application_execution_time +=
+          current_application_execution_time.count();
+      auto submit_time = event.get_profiling_info<
```
DuncanMcBain: You can construct std::chrono::duration instances from these times here and specify that they are nanoseconds. Then later you can easily transform them into milis, or whatever timing you need, using the std::chrono utilities (not least because the values returned from the get_profiling_info call are int64_ts, not floating point, which I think your calculation of current_submission_time will result in).

samples/tiled-convolution/tiled-convolution.cpp
```diff
@@ -0,0 +1,310 @@
...
+  }
+// this part is used to profile the total execution time
+#ifdef PROFILE_SYCL
+  avarage_submission_time =
+      total_submission_time / float(num_host_tile_n * num_host_tile_m);
+  avarage_execution_time =
+      total_execution_time / float(num_host_tile_n * num_host_tile_m);
+  avarage_application_execution_time = total_application_execution_time /
+                                       float(num_host_tile_n * num_host_tile_m);
+  std::cout << " local_size_0, " << ocl_config_t::local_size_m
```
DuncanMcBain: This is also just very untidy. I understand we want to show how to profile your SYCL applications, but I don't think this is the way to do it.

samples/tiled-convolution/tiled-convolution.cpp
```diff
@@ -0,0 +1,310 @@
...
+            << " , total_kernel_execution_time, " << total_execution_time
+            << " , avarage_kernel_submission_time, " << avarage_submission_time
+            << " , avarage_kernel_execution_time, " << avarage_execution_time
+            << " , total_application_execution_time, "
+            << total_application_execution_time
+            << " , avarage_application_execution_time, "
+            << avarage_application_execution_time << "\n";
+#endif
+  // check the correctness of the result
+  auto out_data = out_buff.get_access<read_t>();
```
DuncanMcBain: It won't really matter in this instance, but for consistency this should be in a new scope of its own.

samples/tiled-convolution/tiled-convolution.cpp
```diff
@@ -0,0 +1,310 @@
...
+  for (index_t m = 0; m < total_buffer.m; m++) {
+    for (index_t n = 0; n < total_buffer.m; n++) {
+      if (!(std::abs(index_t(out_data[m][n] - (input_data * filter_data))) <
+            1e-4)) {
+        std::cout << " The result is wrong " << std::endl;
+        return -1;
```
DuncanMcBain: Super pedantic, but convention is to return positive error codes: 1 as a general "went wrong" is a fine error code to return.

samples/tiled-convolution/tiled-convolution.cpp
```diff
@@ -0,0 +1,310 @@
...
+            << " , avarage_kernel_execution_time, " << avarage_execution_time
+            << " , total_application_execution_time, "
+            << total_application_execution_time
+            << " , avarage_application_execution_time, "
+            << avarage_application_execution_time << "\n";
+#endif
+  // check the correctness of the result
+  auto out_data = out_buff.get_access<read_t>();
+  for (index_t m = 0; m < total_buffer.m; m++) {
+    for (index_t n = 0; n < total_buffer.m; n++) {
```
DuncanMcBain: It might be worth using the flat size of the buffer here, because this looks like a mistake: you are referring to total_buffer.m twice, instead of total_buffer.n the second time (at a guess).

samples/tiled-convolution/tiled-convolution.cpp
```diff
@@ -0,0 +1,310 @@
...
+      if (!(std::abs(index_t(out_data[m][n] - (input_data * filter_data))) <
```
DuncanMcBain: I think here you want `std::fabs` instead of a cast (say the result of `out_data[m][n] - (input_data * filter_data` was 0.9; the cast would truncate this to zero which will then pass the condition, even though it was 4 orders of magnitude out).

samples/tiled-convolution/tiled-convolution.cpp
```diff
@@ -0,0 +1,310 @@
...
+    index_t fil_m;                         // filter index m
+    index_t n;                             // input index n
+    index_t fil_n;                         // filter index n
+    typename write_accessor_t::value_type val = 0.0;
+    // loop over filter size m and add halo to input m
+    for (fil_m = 0, m = -1; fil_m < fil_size.m; fil_m++, m++) {
+      index_t in_id_m = (id_m + m >= 0) ? id_m + m : 0;
+      in_id_m = (in_id_m < total_size.m) ? in_id_m : total_size.m - 1;
+      // loop over filter size n and add halo to input n
+      for (fil_n = 0, n = -1; fil_n < 3; fil_n++, n++) {
```
DuncanMcBain: Is this supposed to be fil_size.n instead of 3? It seems weird that it wouldn't match the previous loop.

samples/tiled-convolution/tiled-convolution.cpp
```diff
@@ -0,0 +1,310 @@
...
+  // mask filter value
+  auto filter_data = data_t(0.3);
+  // input array
+  std::vector<data_t> input(total_buffer.size(), input_data);
+  // mask array
+  std::vector<data_t> filter(fil_size.size(), filter_data);
+
+  auto property_list =
+      cl::sycl::property_list{cl::sycl::property::queue::enable_profiling()};
+  // constructing a SYCL queue for OpenCL device where
```
DuncanMcBain: I don't think this comment is useful either - as a longer sample, I think we should be teaching about the algorithms. We have some more basic samples that talk about queues and what they do.

samples/tiled-convolution/tiled-convolution.cpp
```diff
@@ -0,0 +1,310 @@
...
+  // automatically build the underlying context and command_queue for the
+  // chosen device.
+  auto sycl_queue = cl::sycl::queue(
+      (cl::sycl::default_selector()),
```
DuncanMcBain: There should be a queue constructor that takes an async_handler and a property list that doesn't need a device selector. It will simply use that default selector in this case, so you don't need to pass this in.

DuncanMcBain: In fact, it might be best to define this out-of-line of the constructor call, just to keep the indentation lower, but mostly because of the final parameter (the property list). This isn't super necessary though.

DuncanMcBain:
Ah. I see that my old comments are still there. Some of them are comments on code that's subsequently been changed - ignore the ones that don't apply.

samples/tiled-convolution/tiled-convolution.cpp
```diff
@@ -0,0 +1,307 @@
...
+// configuring the openCL local size
+template <typename index_t>
+struct opencl_configuration_t {
+  static constexpr index_t local_size_m = 1;
+  static constexpr index_t local_size_n = 32;
+};
+
+// define the 2-d matrix size
+template <typename index_t>
+class matrix_size_t {
```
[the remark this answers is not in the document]
DuncanMcBain: Discussed offline: they need to be signed, and ranges use unsigned types.

samples/tiled-convolution/tiled-convolution.cpp
```diff
@@ -160,25 +208,22 @@ int main() {
...
   auto property_list =
       cl::sycl::property_list{cl::sycl::property::queue::enable_profiling()};
-  // constructing a SYCL queue for OpenCL device where
-  // automatically build the underlying context and command_queue for the
-  // chosen device.
+
+  auto selector = cl::sycl::default_selector();
   auto sycl_queue = cl::sycl::queue(
-      (cl::sycl::default_selector()),
+      selector,
```
DuncanMcBain: What I was saying with this comment is that you shouldn't need the selector at all, you can remove this and line 215, according to the spec.

samples/tiled-convolution/tiled-convolution.cpp
```diff
@@ -242,63 +279,18 @@ int main() {
...
-            << avarage_application_execution_time << "\n";
+  profiler(events, starts);
 #endif
   // check the correctness of the result
   auto out_data = out_buff.get_access<read_t>();
   for (index_t m = 0; m < total_buffer.m; m++) {
-    for (index_t n = 0; n < total_buffer.m; n++) {
-      if (!(std::abs(index_t(out_data[m][n] - (input_data * filter_data))) <
+    for (index_t n = 0; n < total_buffer.n; n++) {
+      if (!(std::fabs(index_t(out_data[m][n] - (input_data * filter_data))) <
```
DuncanMcBain: You still have the cast to index_t around the expression, this will still likely truncate to zero and lose the erroneous values.

samples/tiled-convolution/tiled-convolution.cpp
```diff
@@ -123,23 +122,72 @@ void inline compute_index(const index_t total_size_dim,
   }
 }
 
+template <typename event_t, typename start_t>
+void inline profiler(event_t& events, const start_t& starts) {
```
DuncanMcBain: Isn't this just std::vector<event_t>? Why is it a template when we know what it is?

DuncanMcBain:
The travis build is still failing. I think this is because std::fabs is in cmath. Also, the change I threatened has been made. If the tests pass, you can make this change, then we can squash it so it doesn't break on the master branch.

samples/tiled-convolution/tiled-convolution.cpp
mehdi-goli: fixed

samples/tiled-convolution/tiled-convolution.cpp
mehdi-goli: fixed

DuncanMcBain:
Have you merged master into this? If so, you'll need to change your CMakeLists:
```CMake
add_sycl_to_target(
TARGET <target name>
SOURCES <sourcefile.cpp>)
```
Which is to say, they're now named parameters and the binary directory parameter has been removed. It's a simple enough change, but a little annoying :)

PR #124, opened 2018-07-03
making the local size to be parametric.

mehdi-goli:


Ruyk:
This looks good to me, simple change and default configuration passes travis.

PR #135, opened 2018-08-24
Update makefile example for PkgConfig support

mmha:
Now that ComputeCpp ships with a `.pc` file the Makefile example should make use of that.

### DIFF
--- a/tools/makefile_basic/Makefile
+++ b/tools/makefile_basic/Makefile
@@ -1,34 +1,28 @@
-# Example Makefile to build a SYCL application using ComputeCpp.
+CXXFLAGS += -std=c++11 `pkg-config --cflags computecpp`
+LDFLAGS += `pkg-config --libs computecpp`
+SYCLFLAGS += `pkg-config computecpp --variable device_cflags`
+SYCLCXX ?= `pkg-config computecpp --variable device_compiler`
 
-# Your ComputeCpp installation root.
-COMPUTECPP_PREFIX ?= /usr/local
+EXECUTABLE = app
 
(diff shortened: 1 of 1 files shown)


mmha:
Yes and no. I extended the help message a bit for that.

- If computecpp is just installed to the root directory nothing needs to be done and everything should Just Work(tm).
- If computecpp is installed to another prefix (more likely), the `.pc` file isn't in a directory that is usually searched by PkgConfig. The user has two possibilities now:
    - Copy the `.pc` file to `/usr/lib/PkgConfig`. No variable needs to be set to build.
    - Add ComputeCpp to the `PKG_CONFIG_PATH` variable, e. g. `export PKG_CONFIG_PATH=<computecpp prefix>/lib/pkgconfig:$PKG_CONFIG_PATH`. This is pretty much the old workflow with a different variable name.

tools/makefile_basic/Makefile
```diff
@@ -1,34 +1,28 @@
...
+
+% : %.o
+	$(CXX) $< $(LDFLAGS) -o $@
 
 .PHONY: clean help
+
 clean:
-	rm -fv $(target) $(target).sycl $(target).o
+	rm -f *.o
+	rm -f $(EXECUTABLE)
```
DuncanMcBain: Can't we leave this as one line? It's also missing the .sycl file, which will be left as part of the compiler output.

mmha: You're right, I forgot the `.sycl` files. I reduced it to one line again.

PR #143, opened 2018-09-11
Update depfile generation for device compiler

DuncanMcBain:
Ninja has a problem if the name of the object doesn't match what it
expects, so this changes the generation to output a relative path, not
an absolute one. Additionally, the quotes in the optional DEPFILE
variable were causing problems, so they have also been removed.


DuncanMcBain:
@mmha you might be interested in this!

PR #144, opened 2018-09-20
Simplify cross compilation with ComputeCpp_HOST_DIR

ProGTX:
Adds a CMake option `ComputeCpp_NATIVE_DIR`  that is used to specify the directory of the ComputeCpp package built for the native architecture. For example, when cross compiling for ARM on an x86 machine, `ComputeCpp_DIR` would point to the ARM package of ComputeCpp and `ComputeCpp_NATIVE_DIR` would point to the x86 one. This is required because the CMake script executes binaries from the ComputeCpp package.

### DIFF
--- a/cmake/Modules/FindComputeCpp.cmake
+++ b/cmake/Modules/FindComputeCpp.cmake
@@ -65,12 +65,23 @@ elseif(DEFINED ENV{COMPUTECPP_DIR})
   set(computecpp_find_hint $ENV{COMPUTECPP_DIR})
 endif()
 
+if(CMAKE_CROSSCOMPILING)
+  # ComputeCpp_NATIVE_DIR is used to find executables that are run locally
+  if(DEFINED ComputeCpp_NATIVE_DIR)
+    set(computecpp_native_find_hint ${ComputeCpp_NATIVE_DIR})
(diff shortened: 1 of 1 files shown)


mmha:
Bikeshed: I think `ComputeCpp_HOST_DIR` is clearer.

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -58,19 +58,29 @@ find_package(OpenCL REQUIRED)
...
 # Find ComputeCpp package
 
-# Try to read the environment variable
 if(DEFINED ComputeCpp_DIR)
   set(computecpp_find_hint ${ComputeCpp_DIR})
 elseif(DEFINED ENV{COMPUTECPP_DIR})
   set(computecpp_find_hint $ENV{COMPUTECPP_DIR})
 endif()
 
+if(CMAKE_CROSSCOMPILING)
```
DuncanMcBain: Argh, I asked you to put this stuff inside CMAKE_CROSSCOMPILING, but that breaks the script (your local test worked because of cache variables). You should initialise the computecpp_native_find_hint to computecpp_find_hint, then let it be mutated in the branch. My fault, sorry!

PR #146, opened 2018-09-27
Check if ComputeCpp target exists before adding

jwlawson:
Rather than just adding the import ComputeCpp target, check to ensure
that the target doesn't already exist. This could happen if a single
cmake list ends up calling `find_package(ComputeCpp)` multiple times.


DuncanMcBain:
Thanks for this! @AdamHarries you might be interested in this :)

DuncanMcBain:
The travis build passed, fwiw

PR #148, opened 2018-10-15
Using Tile convolution with onchip memory and context bound buffer

mehdi-goli:
Adding context bound properties to the buffers in Tiled convolution example; handling compile time local size in Tile convolution CMAKE file.


samples/tiled-convolution/tiled-convolution.cpp
```diff
@@ -32,12 +32,7 @@
 #include <ctime>
 #include <iostream>
 
-#ifndef LOCAL_SIZE_M
-#define LOCAL_SIZE_M 16
-#endif
-#ifndef LOCAL_SIZE_N
-#define LOCAL_SIZE_N 16
-#endif
+// The LOCAL_SIZE_N and LOCAL_SIZE_M is set at compile time
```
Ruyk: is set -> are set

samples/tiled-convolution/CMakeLists.txt
```diff
@@ -1,4 +1,28 @@
...
+    set(LOCAL_SIZE_M 16)
+  endif()
+
+  if (NOT DEFINED LOCAL_SIZE_N)
+    set(LOCAL_SIZE_N 16)
+  endif()
+
+endif()
+
+add_definitions(-DLOCAL_SIZE_M=${LOCAL_SIZE_M})
```
mmha: Can you replace both `add_definitions()` with a single `target_compile_definitions(${SOURCE_NAME} PRIVATE LOCAL_SIZE_M=${LOCAL_SIZE_M} LOCAL_SIZE_N=${LOCAL_SIZE_N})` please?

samples/tiled-convolution/CMakeLists.txt
```diff
@@ -1,4 +1,28 @@
 set(SOURCE_NAME "tiled-convolution")
+
+if(DEFINED TARGET)
```
mmha: You're missing out on a case here: The user might have defined `TARGET` to something other than `RCAR`. In this case the `LOCAL_SIZE_x` variables are never set.

I don't think you need the surrounding check at all.

mmha: Also, is `TARGET` used elsewhere already? If not, I'd prefer a more descriptive name. `TARGET` can mean a lot in build systems.

DuncanMcBain:
OK so somebody broke the Travis config for this repo (it could have been, but was not necessarily, me). Does this continue to pass all tests?

samples/tiled-convolution/CMakeLists.txt
```diff
@@ -14,4 +38,4 @@ add_test(
 install(
   TARGETS ${SOURCE_NAME}
   RUNTIME DESTINATION bin
-)
+)
```
DuncanMcBain: You've removed the newline from this file

samples/tiled-convolution/tiled-convolution.cpp
DuncanMcBain: Why remove the ifndef checks? If you try to compile "manually", this will now stop working

samples/tiled-convolution/tiled-convolution.cpp
```diff
@@ -302,4 +300,4 @@ int main() {
   }
   std::cout << " The result is correct " << std::endl;
   return 0;
-}
+}
```
DuncanMcBain: Newline removed from file

samples/tiled-convolution/CMakeLists.txt
```diff
@@ -1,4 +1,28 @@
 set(SOURCE_NAME "tiled-convolution")
+
```
DuncanMcBain: I think, given all these comments, I'd prefer that instead of all these changes, we unconditionally do what @mmha suggests:
```CMake
target_compile_definitions(
  ${SOURCE_NAME} PRIVATE LOCAL_SIZE_M=1 LOCAL_SIZE_N=32)
```
We can the complexity later - but I think YAGNI

DuncanMcBain:
Hi, I see you've posted more commits, but these seem way beyond the scope of the context_bound property. There's about 1k lines in this PR now - what changes are these? What's the plan with this PR?

DuncanMcBain:
Hi, I see you've posted more commits, but these seem way beyond the scope of the context_bound property. There's about 1k lines in this PR now - what changes are these? What's the plan with this PR?

DuncanMcBain:
This MR has become really noisy, somehow you have a whole bunch of unrelated changes in the diff now (basically every commit to master since October or thereabouts). Can we maybe try to clean this up? It's hard to see what's changed. @mehdi-goli @ProGTX

ProGTX:
I merged master into the branch before creating the PR into @mehdi-goli 's branch, and it seems that the PR merge squashed everything into a single commit, leading to noise. I believe @mehdi-goli fixed it now.

keryell:
I have the feeling that the code would be clearer by using hierarchical parallelism with `parallel_for_work_group` instead of old-style `parallel_for` with `nd_range`.
There would be no need for barriers and the code would be probably more efficient on CPU (actually a lot more efficient with triSYCL implementation).

DuncanMcBain:
I think adding hierarchical parallelism is out of the scope of this MR - we have plenty of other hierarchical elsewhere in this repository. Additionally, I think the CPU is not such an interesting optimisation target for a convolution operation.

PR #149, opened 2018-10-25
MSVC STL version check

MathiasMagnus:
Most of the aggravation of using ComputeCpp with Visual C++ stems from the fact that the STL of MSVC is tested only against latest stable and devel Clang. The check (which is a hard `#error`) can be disabled with a macro.

The addition in this PR tries to ease the life of users in the hiatus periods, when Compute++ has not yet caught up to latest Clang while MSVC already updated their STL headers. It invokes the device compiler on a minimal source file and checks whether the device compiler fails on the version check. If it does, it tries going around the check. It doesn't do elaborate testing of some feature missing from the device compiler. If it fails compiling even after turning off the check, issues a message saying most likely no STL feature will be available. This may even be a `message(FATAL_ERROR)`. Without the STL, not many libraries compile. I made it a message not to exclude users with 100% self-hosted code bases.

### Note

The use of `execute_process` is extremely brittle. I wasted an entire day until I uncovered the problem. Passing of the arguments to `execute_process` is done `VERBATIM`, which goes crazy over the mixed nature of `COMPUTECPP_DEVICE_COMPILER_FLAGS`. Namely its value after invoking computecpp_info.exe is
```
-O2 -mllvm -inline-threshold=1000 -sycl -intelspirmetadata;-sycl-target;spir64
```
which will result in the error
```
error: invalid integral value '2 -mllvm -inline-threshold=1000 -sycl -intelspirmetadata' in '-O2 -mllvm -inline-threshold=1000 -sycl -intelspirmetadata'
```
I would advise against trying to come up with something smarter, that is unless you got some spare time to burn.


MathiasMagnus:
Because I saw that the changes resulted in broken makefiles at times, I revisited the controversial force-include flag generation for MSVC. I tested the more simpler approach with Ninja, NMake Makefiles and MSBuild (the Visual Studio family generators) and all of them work now with this simplified approach.

DuncanMcBain:
Hi @MathiasMagnus, sorry it's taken me so long to review this, I've been working on other projects. This is a great heads-up as our developers (some of whom use Visual Studio) weren't aware of this policy! (Abstractly, I'd heard of it, but I didn't realise they were actually doing hard version checks).

I will take a look but I think my main recommendation will be that, if possible, we put the MSVC-specific stuff in a separate file - but I'll leave that for the review.

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -375,15 +418,9 @@ function(__build_ir)
...
-    # hence the different argument syntax
-    if(CMAKE_GENERATOR MATCHES "Visual Studio")
-      set(forceIncludeFlags "/FI\"${includedFile}\" /TP")
-    else()
-      set(forceIncludeFlags /FI ${includedFile} /TP)
-    endif()
+    set(forceIncludeFlags "/FI ${includedFile} /TP")
   else()
-      set(forceIncludeFlags "-include ${includedFile} -x c++")
+    set(forceIncludeFlags "-include ${includedFile} -x c++")
```
DuncanMcBain: This is good, thanks for doing this

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -131,14 +131,57 @@ execute_process(COMMAND ${ComputeCpp_INFO_EXECUTABLE}
   "--dump-device-compiler-flags"
   OUTPUT_VARIABLE COMPUTECPP_DEVICE_COMPILER_FLAGS
   RESULT_VARIABLE ComputeCpp_INFO_EXECUTABLE_RESULT OUTPUT_STRIP_TRAILING_WHITESPACE)
-list(APPEND COMPUTECPP_DEVICE_COMPILER_FLAGS "-sycl-target ${COMPUTECPP_BITCODE}")
+# convert to list before appending
```
DuncanMcBain: I'm going to be really annoying here, because this wasn't as clean in the upstream as it should be - but I'm not keen on this method of doing things, I think it's going to cause some troubles. I have instead written what I want this code to look like (it causes some changes later in the file. Broadly, the idea is to listify the result from `computecpp_info`, then only append strings to that list. This will maintain the variable in the format we want.

```diff
diff --git a/cmake/Modules/FindComputeCpp.cmake b/cmake/Modules/FindComputeCpp.cmake
index 2f9335a..819df84 100644
--- a/cmake/Modules/FindComputeCpp.cmake
+++ b/cmake/Modules/FindComputeCpp.cmake
@@ -131,13 +131,11 @@ execute_process(COMMAND ${ComputeCpp_INFO_EXECUTABLE}
   "--dump-device-compiler-flags"
   OUTPUT_VARIABLE COMPUTECPP_DEVICE_COMPILER_FLAGS
   RESULT_VARIABLE ComputeCpp_INFO_EXECUTABLE_RESULT OUTPUT_STRIP_TRAILING_WHITESPACE)
-list(APPEND COMPUTECPP_DEVICE_COMPILER_FLAGS "-sycl-target ${COMPUTECPP_BITCODE}")
-
 if(NOT ComputeCpp_INFO_EXECUTABLE_RESULT EQUAL "0")
   message(FATAL_ERROR "compute++ flags - Error obtaining compute++ flags!")
-else()
-  mark_as_advanced(COMPUTECPP_COMPILER_FLAGS)
 endif()
+mark_as_advanced(COMPUTECPP_DEVICE_COMPILER_FLAGS)
+separate_arguments(COMPUTECPP_DEVICE_COMPILER_FLAGS)
 
 find_package_handle_standard_args(ComputeCpp
   REQUIRED_VARS ComputeCpp_ROOT_DIR
@@ -167,8 +165,8 @@ if(CMAKE_CROSSCOMPILING)
   list(APPEND COMPUTECPP_DEVICE_COMPILER_FLAGS -target ${SDK_TARGET_TRIPLE})
 endif()
 
-separate_arguments(COMPUTECPP_DEVICE_COMPILER_FLAGS)
-list(REMOVE_ITEM COMPUTECPP_DEVICE_COMPILER_FLAGS "-emit-llvm")
+list(APPEND COMPUTECPP_DEVICE_COMPILER_FLAGS -sycl-target ${COMPUTECPP_BITCODE})
+list(REMOVE_ITEM COMPUTECPP_DEVICE_COMPILER_FLAGS -emit-llvm)
 message(STATUS "compute++ flags - ${COMPUTECPP_DEVICE_COMPILER_FLAGS}")
 
 if(NOT TARGET OpenCL::OpenCL)
@@ -279,13 +277,11 @@ function(__build_ir)
     list(APPEND target_compile_flags ${source_compile_flags})
   endif()
 
-  set(COMPUTECPP_DEVICE_COMPILER_FLAGS
+  list(APPEND COMPUTECPP_DEVICE_COMPILER_FLAGS
     ${device_compiler_cxx_standard}
-    ${COMPUTECPP_DEVICE_COMPILER_FLAGS}
     ${COMPUTECPP_USER_FLAGS}
     ${target_compile_flags}
   )
-  separate_arguments(COMPUTECPP_DEVICE_COMPILER_FLAGS)
 
   set(ir_dependencies ${SDK_BUILD_IR_SOURCE})
   get_target_property(target_libraries ${SDK_BUILD_IR_TARGET} LINK_LIBRARIES)
```
This patch will apply to the upstream version of FindComputeCpp, but I don't think it will conflict anywhere except these lines.

Edit: OK, suggestions work well for single lines, not so much for entire patches.

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -131,14 +131,57 @@ execute_process(COMMAND ${ComputeCpp_INFO_EXECUTABLE}
...
+# (VERBATIM arg passing to execute_process will go crazy over mixed string-list syntax)
+string(REPLACE " " ";" COMPUTECPP_DEVICE_COMPILER_FLAGS ${COMPUTECPP_DEVICE_COMPILER_FLAGS})
+list(APPEND COMPUTECPP_DEVICE_COMPILER_FLAGS} -sycl-target ${COMPUTECPP_BITCODE})
```
DuncanMcBain: If you use the patch above, this won't be necessary, the comments aren't really needed either (should be obvious that we're operating on a list, and keeping it as a list).

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -131,14 +131,57 @@ execute_process(COMMAND ${ComputeCpp_INFO_EXECUTABLE}
...
   message(FATAL_ERROR "compute++ flags - Error obtaining compute++ flags!")
 else()
   mark_as_advanced(COMPUTECPP_COMPILER_FLAGS)
 endif()
 
+# Check if the hosted STL of MSVC is compatible with ComputeCpp
+if(MSVC)
+  set(ComputeCpp_STL_CHECK_SRC __STL_check)
+  file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/${ComputeCpp_STL_CHECK_SRC}.cpp
+    "#include <ios>\n"
```
DuncanMcBain: Isn't this an internal header - normally, I'd expect <iostream> or similar

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -131,14 +131,57 @@ execute_process(COMMAND ${ComputeCpp_INFO_EXECUTABLE}
...
+              -isystem ${ComputeCpp_INCLUDE_DIRS}
+              -o ${ComputeCpp_STL_CHECK_SRC}.cpp.sycl
+              -c ${ComputeCpp_STL_CHECK_SRC}.cpp
+      WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+      RESULT_VARIABLE ComputeCpp_STL_CHECK_RESULT
+      ERROR_QUIET
+      OUTPUT_QUIET)
+    if(NOT ${ComputeCpp_STL_CHECK_RESULT} EQUAL 0)
+      message(STATUS "Device compiler cannot consume hosted STL headers. Using any parts of the STL will likely result in device compiler errors.")
+	else()
```
DuncanMcBain: I think the formatting has gone a little strange here (maybe just one outdent?)

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -131,14 +131,57 @@ execute_process(COMMAND ${ComputeCpp_INFO_EXECUTABLE}
...
+string(REPLACE " " ";" COMPUTECPP_DEVICE_COMPILER_FLAGS ${COMPUTECPP_DEVICE_COMPILER_FLAGS})
+list(APPEND COMPUTECPP_DEVICE_COMPILER_FLAGS} -sycl-target ${COMPUTECPP_BITCODE})
 
 if(NOT ComputeCpp_INFO_EXECUTABLE_RESULT EQUAL "0")
   message(FATAL_ERROR "compute++ flags - Error obtaining compute++ flags!")
 else()
   mark_as_advanced(COMPUTECPP_COMPILER_FLAGS)
 endif()
 
+# Check if the hosted STL of MSVC is compatible with ComputeCpp
```
DuncanMcBain: Since this whole code section is MSVC-specific, I think I'd be happier if it was part of an opt-in module (i.e. its own file). I feel like that would be a better reflection of the separation of concerns here - it could be called MSVCVersionCheck or similar, and that way users targetting '17 know that they have to deal with it (similarly, once the compiler is updated, they can remove it again).

Let me know what you think!

MathiasMagnus: I'm generally okay with all this being moved to a file of its own. I'm not sure how many such changes are needed. There is a brief GCC-related test at the beginning, which for the sake of consistency might also be put to a different file... all in all, factoring out compiler-specific stuff to separate files and keeping `FindComputeCpp.cmake` to the bare minimum of detecting the library might be the best way to go. Also, checks about the executing runtime mentioned in #155 might also be put to a file of its own.

All of this might be the topic of another PR.

cmake/Modules/FindComputeCpp.cmake
MathiasMagnus: Yes, this should be fixed with a commit.

cmake/Modules/FindComputeCpp.cmake
MathiasMagnus: I don't know why it did not work before. Might have been an MSBuild bug. It all works now.

cmake/Modules/FindComputeCpp.cmake
MathiasMagnus: I'm not good at reading diff files, but I believe I made the same changes to my working tree. It seems to work.

cmake/Modules/FindComputeCpp.cmake
MathiasMagnus: See before

cmake/Modules/FindComputeCpp.cmake
DuncanMcBain: Yes, you have. Thanks for this, something felt "off" about it for ages, but your changes spurred me into making the right change. Sometimes it's good to prune back old bits!

cmake/Modules/FindComputeCpp.cmake
DuncanMcBain: I think I'd prefer if we move it into its own file for now, provided it's not *too* difficult. Let me know if it looks like it will actually be too onerous.

You're right about the gcc/clang check. Revisiting it, I'm tempted to remove it, because who uses 4.8? :stuck_out_tongue_closed_eyes: More seriously, if you use old compilers like that, you will get compile issues. Though maybe we should move it to this new file while we still (ostensibly) support Ubuntu 14.04?

Decision time: if you feel like moving it too, please feel free, though I won't mark it as a requirement of this PR by any means. Thanks again!

PR #150, opened 2018-10-29
Add check so `end` iterator is not dereferenced

jwlawson:
The current behaviour of `get_node` in the virtual pointer mapper will
use `lower_bound` to get the buffer which is not less than the input
pointer. If the pointer mapper has only allocated one buffer, and we are
looking up a pointer offset into that buffer then the call to
`lower_bound` will return an iterator pointing past the end of the map.
This means we cannot dereference this iterator, as it does not point
into the map, but we still want to decrement this iterator to get the
buffer that the virtual pointer is referencing.

By adding a check for whether the iterator is the end we can short
circuit the iterator dereference, and so avoid this problem without
otherwise changing the behaviour of the pointer mapper.

### DIFF
--- a/include/vptr/virtual_ptr.hpp
+++ b/include/vptr/virtual_ptr.hpp
@@ -219,7 +219,7 @@ class PointerMapper {
     auto node = m_pointerMap.lower_bound(ptr);
     // If the value of the pointer is not the one of the node
     // then we return the previous one
-    if (node->first != ptr) {
+    if (node == m_pointerMap.end() || node->first != ptr) {
       if (node == std::begin(m_pointerMap)) {
         throw std::out_of_range("The pointer is not registered in the map");
       }


jwlawson:
The current tests exhibit this behaviour, but silently pass until you turn on debug iterators (eg. using `-D_GLIBCXX_DEBUG`).

I'm not sure what a more explicit test would look like.

DuncanMcBain:
@jwlawson is correct - the first time we tried the SDK on Windows after the virtual pointer was added, it failed because Windows debug builds use checked iterators by default.

Unfortunately the _GLIBCXX_DEBUG isn't totally safe - you can't pass containers between TUs that have DEBUG and not DEBUG, so I'm not sure you could reliably create a test that links against a non-_GLIBCXX_DEBUG ComputeCpp. It *might* work if you do target_definitions() on the tests... The test "offset" fails when built in debug mode. If it definitely doesn't share containers with the library, it should work!

PR #152, opened 2018-11-05
Fix type shadowing

Rbiessy:
The type `buffer_t` is defined in the class `PointerMapper` and is redefined to a different type in `get_buffer` which gives shadowing warnings.
`SYCLmalloc` doesn't have this issue but the alias should probably be changed for consistency.
The naming isn't great, renaming the first `buffer_t` to `buffer_mem_t` would make more sense but it is used in more places.


DuncanMcBain:
I'm not opposed to larger changes if you think they would be better, personally. Thanks for this fix. Let me know if you want to make a different change.

Rbiessy:
I personally prefer this version.

PR #157, opened 2018-11-23
Fix accessor tests, and add property_list parameter to syclmalloc

AdamHarries:
This PR fixes a test that was forgotten when travis broke, and adds a `parameter_list` argument to syclmalloc so that we can control allocation (e.g. `use_host_ptr`, `context_bound` etc).

### DIFF
--- a/tests/vptr/accessor.cc
+++ b/tests/vptr/accessor.cc
@@ -121,9 +121,8 @@ TEST(accessor, two_buffers) {
   }
 }
 
+// Is this test still required, given that we no longer use allocators?
 TEST(accessor, allocator) {
-  // an allocator type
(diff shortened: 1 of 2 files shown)


tests/vptr/accessor.cc
```diff
@@ -121,9 +121,8 @@ TEST(accessor, two_buffers) {
   }
 }
 
+// Is this test still required, given that we no longer use allocators?
```
AdamHarries: I'm not sure if this test is still required, as as far as I can tell, all it does is test that we've actually added two buffers?

mehdi-goli: No, Thanks for fixing it

PR #158, opened 2018-11-30
Added initial example of builtin kernel

Ruyk:
This patch adds an example usage of the Codeplay extension CP018 (https://github.com/codeplaysoftware/standards-proposals/blob/master/builtin_kernels/index.md) , which will be part of the upcoming 1.0.3 release.

### DIFF
--- a/samples/builtin-kernel-example/builtin-kernel-example.cpp
+++ b/samples/builtin-kernel-example/builtin-kernel-example.cpp
@@ -1,6 +1,6 @@
 /***************************************************************************
  *
- *  Copyright (C) 2016 Codeplay Software Limited
+ *  Copyright (C) 2018 Codeplay Software Limited
  *  Licensed under the Apache License, Version 2.0 (the "License");
  *  you may not use this file except in compliance with the License.
  *  You may obtain a copy of the License at
@@ -31,9 +31,11 @@
 #include <array>
 #include <iostream>
 
+using namespace cl::sycl;
+
 int main() {
 
-  queue testQueue(TEST_SELECTOR, TEST_ASYNC_HANDLER);
+  queue testQueue;
   context testContext = testQueue.get_context();
 
   device dev = testContext.get_devices()[0];
--- a/samples/builtin-kernel-example/builtin-kernel-example.cpp
+++ b/samples/builtin-kernel-example/builtin-kernel-example.cpp
@@ -1,0 +1,119 @@
+/***************************************************************************
+ *
+ *  Copyright (C) 2016 Codeplay Software Limited
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  For your convenience, a copy of the License has been included in this
+ *  repository.
+ *
+ *  Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


samples/builtin-kernel-example/builtin-kernel-example.cpp
```diff
@@ -0,0 +1,119 @@
...
+ **************************************************************************/
+
+#include <CL/sycl.hpp>
+
+#include <array>
+#include <iostream>
+
+int main() {
+
+  queue testQueue(TEST_SELECTOR, TEST_ASYNC_HANDLER);
```
DuncanMcBain: This is missing `cl::sycl::`, or `using namespace` elsewhere in the file. I think for a sample, either is fine.
This is also using the TEST_SELECTOR stuff, which isn't available in the SDK - probably suffices to replace with a default_selector and a handler that simply throws the first exception, with optional output.

samples/builtin-kernel-example/builtin-kernel-example.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+    testQueue.submit([&](handler& cgh) {
+      auto accIn = buf.get_access<access::mode::read>(cgh);
+      auto accOut = bufOut.get_access<access::mode::write>(cgh);
+
+      /* Using the OpenCL interoperability to set the kernel 
+       * arguments to match the accessors to the builtin kernel
+       * arguments.
+       */
+      cgh.set_arg(0, accIn);
```
DuncanMcBain: Can we use `handler::set_args` here?

samples/builtin-kernel-example/builtin-kernel-example.cpp
```diff
@@ -0,0 +1,119 @@
...
+      return 0;
+  }
+
+  /*
+   * The only builtin kernel supported by this example is the
+   * ComputeAorta copy buffer, defined as:
+   *    copy_buffer(__global * in, __global * out)
+   * And copies input into output.
+   */
+  const std::string kAortaTestKernelName{"copy_buffer"};
```
DuncanMcBain: Not super keen on the `k` for constant naming convention, but it's no big deal. `copy_buffer_kernel` might be more informative (though perhaps you want to preserve the ComputeAorta name?).

samples/builtin-kernel-example/builtin-kernel-example.cpp
Ruyk: That was the wrong commit actually, updated now.

samples/builtin-kernel-example/builtin-kernel-example.cpp
Ruyk: Why? this is the current usage in the SYCL spec

samples/builtin-kernel-example/builtin-kernel-example.cpp
Ruyk: I have to preserve that kernel name or it cannot be tested on Aorta.

samples/builtin-kernel-example/builtin-kernel-example.cpp
DuncanMcBain: It's by no means necessary, but I thought that the varags version would be nicer - no big deal though!

samples/builtin-kernel-example/builtin-kernel-example.cpp
DuncanMcBain: I meant the name of the variable sorry, not the string it is containing - again, not a necessary change!

PR #160, opened 2018-12-06
Move compiler checks to separate module

### LINKED ISSUE #159: Forced include mechanism in FindComputeCpp.cmake does not work in Visual Studio 2015
Forced include method doesn't work properly under VS2015.
The line:
set(forceIncludeFlags "/FI ${includedFile} /TP")
Gets split by spaces and the "includedFile" becomes part of the "AdditionalOptions" and a blank "ForcedIncludeFiles" tag is created.

If the line is changed to:
set(forceIncludeFlags "/FI${includedFile} /TP")
The ForcedInclude tag is properly populated with "includedFile". Note that I haven't tested the solution for paths with spaces or multiple files.

DuncanMcBain:
Moving these compiler checks to a separate module keeps the main
FindComputeCpp.cmake clean and makes the checks "opt-in", if you
know that you won't need the check for your project.

@MathiasMagnus @ProGTX I'd appreciate if you could look at this!
Should fix #159

### DIFF
--- a/cmake/Modules/ComputeCppCompilerChecks.cmake
+++ b/cmake/Modules/ComputeCppCompilerChecks.cmake
@@ -1,0 +1,50 @@
+cmake_minimum_required(VERSION 3.4.3)
+
+if(CMAKE_COMPILER_IS_GNUCXX)
+  if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 4.8)
+    message(FATAL_ERROR "host compiler - gcc version must be > 4.8")
+  endif()
+elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
+  if (${CMAKE_CXX_COMPILER_VERSION} VERSION_LESS 3.6)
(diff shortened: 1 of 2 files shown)


MathiasMagnus:
I'll take a look on monday. I've not checked the code out yet, but I generally agree with the comment.

ProGTX:
Looks good, seems to work.

DuncanMcBain:
I kind of want it to execute just on import, since that's kind of the point. If you want the compiler checks, boom, `import(ComputeCppCompilerChecks)`. If you don't want them... don't :stuck_out_tongue:

MathiasMagnus:
Fair enough… I mean it's just bikeshedding afterall. The functionality is there, if it really hurts anyone, we can reiterate. The checks have to be made anyhow, so this will be invoked inside FindComputeCpp.cmake anyway. Perhaps the nuance of a difference might be apparent when we go about guarding FindComputeCpp.cmake against multiple runs.

Issue #126, opened 2018-07-05
Return an error when ComputeCpp_INFO_EXECUTABLE cannot be found


alistairlow:
Running with ComputeCpp 0.9.0 on 32-bit ARM. If the computecpp_info binary cannot be found, the command silently succeeds, but ${ComputeCpp_INFO_EXECUTABLE} "--dump-version" fails. It would be easier to debug if the initial error was reported.

DuncanMcBain:
Fixed by #161

PR #161, opened 2018-12-07
Report an error when computecpp_info or compute++ can't be found (#126)

RossBrunton:
Added a thing to the makefile, let me know if there is a better way of doing this.

### DIFF
--- a/cmake/Modules/FindComputeCpp.cmake
+++ b/cmake/Modules/FindComputeCpp.cmake
@@ -106,34 +106,38 @@ get_filename_component(computecpp_canonical_root_dir "${ComputeCpp_INCLUDE_DIRS}
 set(ComputeCpp_ROOT_DIR "${computecpp_canonical_root_dir}" CACHE PATH
     "The root of the ComputeCpp install")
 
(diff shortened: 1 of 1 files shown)


Ruyk:
That looks good actually, thanks! I

DuncanMcBain:
I don't really want to add this, because it will break projects with optional ComputeCpp support (you shouldn't need ComputeCpp to be present if you aren't going to use that). This is handled via the `find_package_handle_standard_args`* call in the module - though I can see that if it can't find the executable, that is going to fail anyway. I will think about this.

*When you do `find_package`, if you say `REQUIRED`, then it will bail when it can't find it - otherwise it just moves on with its life, even when not found.

Ruyk:
@DuncanMcBain You are right that REQUIRED should handle that, but the error you get if computecpp_info (at least on my experience) is not great, and difficult to track. This would have a clearer error. Maybe there is a better way to get the clear error? I am not fixated on the solution.
Also, if you don't use ComputeCpp in your project, why would you have FindComputeCpp directly on the CMake? wouldn't you have some if /else around it anyway?

PR #163, opened 2018-12-10
Example of synchronous error handling

### LINKED ISSUE #105: Make sample that has both synchronous and asynchronous exceptions
SYCL code can have both synchronous and asynchronous exceptions. While our samples very well demonstrate async handlers, errors that happen synchronously are exhibited less, and are dealt with even more poorly. To that end we should add a sample that shows some of the ways that synchronous exceptions can be thrown (e.g. a device selector that returns a negative number for all devices, contexts constructed from invalid OpenCL contexts - there are lots of possibilities).

RossBrunton:
This example is specifically about a handler which fails to find a valid
device.

This should fix #105.

----

Once again, let me know if I'm not writing my code "sycl-y" enough, or you want another example with the CL context thing.

### DIFF
--- a/samples/sync-handler/sync-handler.cpp
+++ b/samples/sync-handler/sync-handler.cpp
@@ -1,0 +1,91 @@
+/***************************************************************************
+ *
+ *  Copyright (C) 2018 Codeplay Software Limited
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  For your convenience, a copy of the License has been included in this
+ *  repository.
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *  Codeplay's ComputeCpp SDK
+ *
+ *  sync-handler.cpp
+ *
+ *  Description:
+ *    Sample code that demonstrates the use of a synchronous error handler to
+ *    demonstrate error handling.
+ *
+ **************************************************************************/
+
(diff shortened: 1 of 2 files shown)


samples/sync-handler/sync-handler.cpp
```diff
@@ -0,0 +1,94 @@
...
+    buffer<size_t, 1> buf(sevens.data(), range<1>(number_of_sevens));
+    myQueue.submit([&](handler& cgh) {
+      auto ptr = buf.get_access<access::mode::discard_write>(cgh);
+      cgh.parallel_for<Worker>(range<1>(number_of_sevens),
+                               [=](item<1> item) { ptr[item] = 7; });
+    });
+  } catch (exception const& e) {
+    /* Report the exception to the user. */
+    std::cout << "SYCL exception caught: " << e.what() << "\n";
+    std::cout << "Running on host...\n";
```
cjdb: These ought to be writes to `cerr`, rather than `cout`.

samples/sync-handler/sync-handler.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+    /* If the selector couldn't find a suitable device, the following code
+     * will not run. */
+    buffer<size_t, 1> buf(sevens.data(), range<1>(number_of_sevens));
+    myQueue.submit([&](handler& cgh) {
+      auto ptr = buf.get_access<access::mode::discard_write>(cgh);
+      cgh.parallel_for<Worker>(range<1>(number_of_sevens),
+                               [=](item<1> item) { ptr[item] = 7; });
+    });
+  } catch (exception const& e) {
```
cjdb: Please choose one of east-`const` or `const`-west per file. I recommend east-`const`, since that means that your `const` is always to the right of whatever you want to `const`-qualify.

samples/sync-handler/sync-handler.cpp
```diff
@@ -0,0 +1,91 @@
...
+ **************************************************************************/
+
+#include <CL/sycl.hpp>
+
+#include <iostream>
+
+using namespace cl::sycl;
+
+namespace {
+class Worker;
```
cjdb: This is a declaration. IIRC we need this to be a definition (cc @Ruyk).

PR #170, opened 2019-01-10
Various CMake improvements

Rbiessy:
Various changes mostly needed by Eigen:
- fix issue if multiple targets use the same source file
- ~take into account arguments from the `COMPILE_FLAGS` target property~ Removed as it may break the compiler if some flags are meant for a different host compiler only (i.e. gcc)
- call separate_arguments on `COMPUTECPP_USER_FLAGS` as the user should just be able to give space separated string
- remove unused `device_compiler_includes` variable

### DIFF
--- a/cmake/Modules/FindComputeCpp.cmake
+++ b/cmake/Modules/FindComputeCpp.cmake
@@ -243,13 +243,10 @@ function(__build_ir)
 
   set(include_directories "$<TARGET_PROPERTY:${SDK_BUILD_IR_TARGET},INCLUDE_DIRECTORIES>")
   set(compile_definitions "$<TARGET_PROPERTY:${SDK_BUILD_IR_TARGET},COMPILE_DEFINITIONS>")
-  set(compile_options "$<TARGET_PROPERTY:${SDK_BUILD_IR_TARGET},COMPILE_OPTIONS>")
   set(generated_include_directories
     $<$<BOOL:${include_directories}>:-I\"$<JOIN:${include_directories},\"\t-I\">\">)
   set(generated_compile_definitions
     $<$<BOOL:${compile_definitions}>:-D$<JOIN:${compile_definitions},\t-D>>)
-  set(generated_compile_options
-    $<$<BOOL:${compile_options}>:$<JOIN:${compile_options},\t>>)
 
   # Obtain language standard of the file
   set(device_compiler_cxx_standard)
@@ -303,7 +300,6 @@ function(__build_ir)
     OUTPUT ${outputSyclFile}
     COMMAND ${ComputeCpp_DEVICE_COMPILER_EXECUTABLE}
             ${COMPUTECPP_DEVICE_COMPILER_FLAGS}
-            ${generated_compile_options}
             ${generated_include_directories}
             ${generated_compile_definitions}
             -o ${outputSyclFile}
--- a/cmake/Modules/FindComputeCpp.cmake
+++ b/cmake/Modules/FindComputeCpp.cmake
@@ -266,19 +266,20 @@ function(__build_ir)
     set(device_compiler_cxx_standard "")
   endif()
 
-  list(APPEND COMPUTECPP_DEVICE_COMPILER_FLAGS
-    ${device_compiler_cxx_standard}
-    ${COMPUTECPP_USER_FLAGS}
-  )
-
   get_property(source_compile_flags
     SOURCE ${SDK_BUILD_IR_SOURCE}
     PROPERTY COMPUTECPP_SOURCE_FLAGS
   )
   separate_arguments(source_compile_flags)
   if(source_compile_flags)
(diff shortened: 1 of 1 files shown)


cmake/Modules/FindComputeCpp.cmake
```diff
@@ -259,19 +262,25 @@ function(__build_ir)
...
-    list(APPEND target_compile_flags ${source_compile_flags})
+    list(APPEND COMPUTECPP_DEVICE_COMPILER_FLAGS ${source_compile_flags})
   endif()
 
-  list(APPEND COMPUTECPP_DEVICE_COMPILER_FLAGS
-    ${device_compiler_cxx_standard}
-    ${COMPUTECPP_USER_FLAGS}
-    ${target_compile_flags}
-  )
+  get_target_property(target_compile_flags ${SDK_BUILD_IR_TARGET} COMPILE_FLAGS)
```
jwlawson: Would it be better to use a generator expression to extract the compile flags? That way a user can add flags to the target after calling `add_sycl_to_target` and they will still be added to the custom command.

Rbiessy: That's a good point, it would be good yeah. I'm not too comfortable with generators but I can try that.

Rbiessy: I tried something similar to the 2 generators above but I can't make it work if the property is a space separated string (which it is in Eigen). It is considered as a single argument so the `join` does nothing and once it is given to `add_custom_command` CMake will backslash the spaces...
There is a similar issue here: http://www.gccxml.org/Bug/view.php?id=14353#c33712
The only solution I see is to write all the arguments to a file and tell the compiler to read the arguments from there.
Note that the 2 others generators would not work with a space separated property either they just don't happen to be used in Eigen. Is Eigen in the wrong here or is this something we should support?

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -234,7 +235,9 @@ function(__build_ir)
   get_filename_component(sourceFileName ${SDK_BUILD_IR_SOURCE} NAME)
 
   # Set the path to the integration header.
-  set(outputSyclFile ${CMAKE_CURRENT_BINARY_DIR}/${sourceFileName}.sycl)
+  # The .sycl filename must depend on the target so that different targets
+  # using the same source file will be generated with a different rule.
+  set(outputSyclFile ${CMAKE_CURRENT_BINARY_DIR}/${SDK_BUILD_IR_TARGET}_${sourceFileName}.sycl)
   set(depFileName ${CMAKE_CURRENT_BINARY_DIR}/${sourceFileName}.sycl.d)
```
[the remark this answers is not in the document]
Rbiessy: Will do.

cmake/Modules/FindComputeCpp.cmake
jwlawson: The option should be a semicolon separated cmake list, not space separated. That sounds like an Eigen problem to me.

Also I just noticed this is using `COMPILE_FLAGS` --- this is deprecated and you should be using `COMPILE_OPTIONS`. I have no idea if that works if Eigen is setting `COMPILE_FLAGS` though.

Rbiessy: Ok I could convert the COMPILE_FLAGS string to a COMPILE_OPTIONS list just in Eigen only for SYCL so that's working fine, thanks.

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -234,15 +235,21 @@ function(__build_ir)
...
+  set(depFileName ${baseSyclName}.sycl.d)
 
   set(include_directories "$<TARGET_PROPERTY:${SDK_BUILD_IR_TARGET},INCLUDE_DIRECTORIES>")
   set(compile_definitions "$<TARGET_PROPERTY:${SDK_BUILD_IR_TARGET},COMPILE_DEFINITIONS>")
+  set(compile_options "$<TARGET_PROPERTY:${SDK_BUILD_IR_TARGET},COMPILE_OPTIONS>")
   set(generated_include_directories
     $<$<BOOL:${include_directories}>:-I\"$<JOIN:${include_directories},\"\t-I\">\">)
   set(generated_compile_definitions
     $<$<BOOL:${compile_definitions}>:-D$<JOIN:${compile_definitions},\t-D>>)
+  set(generated_compile_options
```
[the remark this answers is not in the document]
Rbiessy: I tried without it but then `compile_options` is given as is to `add_custom_command` (i.e. as a semicolon separated string). I'm guessing even though `add_custom_command` accepts generators it is not able to automatically convert the list to a space separated string.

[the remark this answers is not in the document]
DuncanMcBain: Roger that!

PR #171, opened 2019-01-10
Add support for compiler driver

Rbiessy:
This allows to build a target using compute++ driver. For now CMAKE_CXX_COMPILER has to be set as late as possible (i.e just before calling add_sycl_to_target) otherwise CMake will reset the cache.
This doesn't affect any of the current samples in this repository.

I am not too sure about the variable naming convention. As far as I understand it we use camelCase for variable that are not exposed?

### DIFF
--- a/cmake/Modules/FindComputeCpp.cmake
+++ b/cmake/Modules/FindComputeCpp.cmake
@@ -94,7 +94,6 @@ set(ComputeCpp_ROOT_DIR "${computecpp_canonical_root_dir}" CACHE PATH
 
 if(NOT ComputeCpp_INFO_EXECUTABLE)
   message(WARNING "Can't find computecpp_info - check ComputeCpp_DIR")
-  set(COMPUTECPP_DEVICE_COMPILER_FLAGS "")
 else()
   execute_process(COMMAND ${ComputeCpp_INFO_EXECUTABLE} "--dump-version"
     OUTPUT_VARIABLE ComputeCpp_VERSION
@@ -116,17 +115,7 @@ else()
       message(WARNING "platform - your system CANNOT support ComputeCpp")
     endif()
   endif()
-
-  execute_process(COMMAND ${ComputeCpp_INFO_EXECUTABLE}
-    "--dump-device-compiler-flags"
-    OUTPUT_VARIABLE COMPUTECPP_DEVICE_COMPILER_FLAGS
-    RESULT_VARIABLE ComputeCpp_INFO_EXECUTABLE_RESULT OUTPUT_STRIP_TRAILING_WHITESPACE)
-  if(NOT ComputeCpp_INFO_EXECUTABLE_RESULT EQUAL "0")
-    message(WARNING "compute++ flags - Error obtaining compute++ flags!")
-  endif()
 endif()
-mark_as_advanced(COMPUTECPP_DEVICE_COMPILER_FLAGS)
-separate_arguments(COMPUTECPP_DEVICE_COMPILER_FLAGS)
 
 find_package_handle_standard_args(ComputeCpp
   REQUIRED_VARS ComputeCpp_ROOT_DIR
@@ -148,6 +137,9 @@ if(NOT ComputeCpp_FOUND)
   return()
 endif()
 
+list(APPEND COMPUTECPP_DEVICE_COMPILER_FLAGS -O2 -mllvm -inline-threshold=1000 -intelspirmetadata)
+mark_as_advanced(COMPUTECPP_DEVICE_COMPILER_FLAGS)
+
 if(CMAKE_CROSSCOMPILING)
   if(NOT COMPUTECPP_DONT_USE_TOOLCHAIN)
     list(APPEND COMPUTECPP_DEVICE_COMPILER_FLAGS --gcc-toolchain=${COMPUTECPP_TOOLCHAIN_DIR})
@@ -157,7 +149,6 @@ if(CMAKE_CROSSCOMPILING)
 endif()
 
 list(APPEND COMPUTECPP_DEVICE_COMPILER_FLAGS -sycl-target ${COMPUTECPP_BITCODE})
-list(REMOVE_ITEM COMPUTECPP_DEVICE_COMPILER_FLAGS -emit-llvm)
 message(STATUS "compute++ flags - ${COMPUTECPP_DEVICE_COMPILER_FLAGS}")
 
 if(NOT TARGET OpenCL::OpenCL)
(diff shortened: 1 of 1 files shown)


cmake/Modules/FindComputeCpp.cmake
```diff
@@ -397,19 +396,44 @@ function(add_sycl_to_target)
     "${multi_value_args}"
     ${ARGN}
   )
-  set(fileCounter 0)
-  # Add custom target to run compute++ and generate the integration header
-  foreach(sourceFile ${SDK_ADD_SYCL_SOURCES})
-    if(NOT IS_ABSOLUTE ${sourceFile})
-      set(sourceFile "${CMAKE_CURRENT_SOURCE_DIR}/${sourceFile}")
+
+  # If the CXX compiler is set to compute++ enable the driver. This removes the
```
DuncanMcBain: Just the first line as a comment is sufficient, I think

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -397,19 +396,44 @@ function(add_sycl_to_target)
...
+  # need to explicitly generate the sycl integration header in a separate
+  # compiler pass.
+  get_filename_component(cmakeCxxCompilerFileName "${CMAKE_CXX_COMPILER}" NAME)
+  if("${cmakeCxxCompilerFileName}" STREQUAL "compute++")
+    if(MSVC)
```
DuncanMcBain: We have a separate CMake file for compiler and platform checks, though I suppose it's no good if you get junk on Windows. What happens? Do you get a sensible error message? Can we use compute++ as a Windows compiler without using sycl-driver?

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -397,19 +396,44 @@ function(add_sycl_to_target)
...
+
+    # Set all flags required to compile the target when using the compiler driver
+    list(APPEND computecppDriverCompileFlags ${COMPUTECPP_DEVICE_COMPILER_FLAGS}
+                                             ${COMPUTECPP_USER_FLAGS})
+    get_target_property(includeAfter ${SDK_ADD_SYCL_TARGET} COMPUTECPP_INCLUDE_AFTER)
+    if(includeAfter)
+      list(APPEND computecppDriverCompileFlags -fsycl-ih-last)
+    endif()
+    # Remove flags that are handled by the driver.
+    list(REMOVE_ITEM computecppDriverCompileFlags -emit-llvm -sycl)
```
DuncanMcBain: We get these flags from the computecpp_info tool, but it's starting to look like we might as well maintain our own list here, since the computecpp_info ones are occasionally outdated (like -emit-llvm). That way, you'd not have to remove any here.

Rbiessy: Yeah I'm not a big fan of the flags given by computecpp_info to be honest.
It would break a lot of existing CMake samples if we stop using them by default though :/

cmake/Modules/FindComputeCpp.cmake
Rbiessy: So I have no idea what the state of compute++ on Windows is today with or without the driver. All I know is that apparently when it was tried with Eigen, driver was not working but 2 steps compiler was. I don't have a way of testing on Windows either :/

cmake/Modules/FindComputeCpp.cmake
Rbiessy: Nevermind ignore me it would still work just fine. We could just by default add `-O2 -mllvm -inline-threshold=1000 -intelspirmetadata` to `COMPUTECPP_DEVICE_COMPILER_FLAGS` and later either add `-sycl-driver` or `-sycl`. Would this mean we would be able to remove the call to `computecpp_info --dump-is-supported` as well as `--dump-device-compiler-flags`?

DuncanMcBain: The is-supported thing is arguably becoming less useful, as it will tell you some platforms aren't supported when they are. And it doesn't make sense when cross-compiling either. And, in fact, I don't know how it determines which platform you're targeting (you might have several, only one of which is supported). It's possible, then, that removing a couple of those calls wouldn't be a bad thing. It makes the followup code so much simpler (`-sycl` can then be added only inside build_ir, for example). I'd like to keep `computecpp_info` as a part of the whole find process, though, users might still want to do whatever with it.

Rbiessy: I only removed the call to --dump-device-compiler-flags as I think --dump-is-supported is unrelated to this PR after all.

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -401,19 +391,51 @@ function(add_sycl_to_target)
...
-    MATH(EXPR fileCounter "${fileCounter} + 1")
-  endforeach()
+
+    get_target_property(includeAfter ${SDK_ADD_SYCL_TARGET} COMPUTECPP_INCLUDE_AFTER)
+    if(includeAfter)
+      list(APPEND COMPUTECPP_USER_FLAGS -fsycl-ih-last)
+    endif()
+    list(INSERT COMPUTECPP_DEVICE_COMPILER_FLAGS 0 -sycl-driver)
+    # Prepend COMPUTECPP_DEVICE_COMPILER_FLAGS and append COMPUTECPP_USER_FLAGS
+    foreach(prop COMPILE_OPTIONS INTERFACE_COMPILE_OPTIONS)
```
Rbiessy: This whole foreach loop is supposed to be equivalent to:
```
target_compile_options(${SDK_ADD_SYCL_TARGET} BEFORE PUBLIC ${COMPUTECPP_DEVICE_COMPILER_FLAGS})
target_compile_options(${SDK_ADD_SYCL_TARGET} PUBLIC ${COMPUTECPP_USER_FLAGS})
```
I don't understand why the `BEFORE` option is not prepending the flags as I think it should. If we don't prepend the default `-O2` could override the user's optimization level.
Should we only add `-O2` for the 2 steps compilation instead?

cmake/Modules/FindComputeCpp.cmake
DuncanMcBain: It is unrelated, but I still wouldn't mind it being removed :p

cmake/Modules/FindComputeCpp.cmake
DuncanMcBain: I don't really mind. `O2` is definitely needed, right enough. Can't you `list(PREPEND COMPUTECPP_USER_FLAGS -O2 etc)`?

Rbiessy: It is an issue if you want to debug and give `O0`.
`list(PREPEND ...)` won't make it simpler, I would still need to do the loop for both `COMPILE_OPTIONS` and `INTERFACE_COMPILE_OPTIONS`, to read the property and to set it again.

PR #177, opened 2019-02-11
Remove codeplay specific implementation details from vptr

DuncanMcBain:


AdamHarries:
LGTM, modulo tests passing on travis. As discussed, we might need to be aware of issues with the reinterpret casting of buffers, and a comment might be useful to warn future users that there might still be bugs, but I imagine that having this in the SDK will just root them out faster, at worst.

Ruyk:
LGTM

PR #182, opened 2019-02-19
Add kernel object file as compilation artifact

Naghasan:
Allow an easier access to the kernel binary for inspection

### DIFF
--- a/cmake/Modules/FindComputeCpp.cmake
+++ b/cmake/Modules/FindComputeCpp.cmake
@@ -31,6 +31,7 @@
 
 cmake_minimum_required(VERSION 3.4.3)
 include(FindPackageHandleStandardArgs)
+include(ComputeCppIRMap)
 
 set(COMPUTECPP_USER_FLAGS "" CACHE STRING "User flags for compute++")
 separate_arguments(COMPUTECPP_USER_FLAGS)
@@ -230,7 +231,7 @@ function(__build_ir)
   # using the same source file will be generated with a different rule.
   set(baseSyclName ${CMAKE_CURRENT_BINARY_DIR}/${SDK_BUILD_IR_TARGET}_${sourceFileName})
   set(outputSyclFile ${baseSyclName}.sycl)
-  set(outputDeviceFile ${baseSyclName}.o)
+  set(outputDeviceFile ${baseSyclName}.${IR_MAP_${COMPUTECPP_BITCODE}})
   set(depFileName ${baseSyclName}.sycl.d)
 
   set(include_directories "$<TARGET_PROPERTY:${SDK_BUILD_IR_TARGET},INCLUDE_DIRECTORIES>")
--- a/cmake/Modules/FindComputeCpp.cmake
+++ b/cmake/Modules/FindComputeCpp.cmake
@@ -230,6 +230,7 @@ function(__build_ir)
   # using the same source file will be generated with a different rule.
   set(baseSyclName ${CMAKE_CURRENT_BINARY_DIR}/${SDK_BUILD_IR_TARGET}_${sourceFileName})
   set(outputSyclFile ${baseSyclName}.sycl)
+  set(outputDeviceFile ${baseSyclName}.o)
   set(depFileName ${baseSyclName}.sycl.d)
 
   set(include_directories "$<TARGET_PROPERTY:${SDK_BUILD_IR_TARGET},INCLUDE_DIRECTORIES>")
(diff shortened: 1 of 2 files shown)


cmake/Modules/FindComputeCpp.cmake
```diff
@@ -308,7 +310,7 @@ function(__build_ir)
 
   if(NOT MSVC)
     # Add a custom target for the generated integration header
-    add_custom_target(${headerTargetName} DEPENDS ${outputSyclFile})
+    add_custom_target(${headerTargetName} DEPENDS ${outputDeviceFile} ${outputSyclFile})
```
Ruyk: I think the documentation at the top of the file should be updated referring to the fact that now we output the bc file as well as the sycl one

Naghasan: good point, fixed now

mmha: The object file isn't really a dependency of the header target.

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -230,6 +230,7 @@ function(__build_ir)
   # using the same source file will be generated with a different rule.
   set(baseSyclName ${CMAKE_CURRENT_BINARY_DIR}/${SDK_BUILD_IR_TARGET}_${sourceFileName})
   set(outputSyclFile ${baseSyclName}.sycl)
+  set(outputDeviceFile ${baseSyclName}.o)
```
DuncanMcBain: Does this output an object file now, or is this a bitcode file? FindComputeCpp.cmake doesn't use -sycl-driver yet (my fault, I never really tried hard to add it in a sensible way)

cmake/Modules/FindComputeCpp.cmake
Naghasan: Kind of, the content of the object file is in the sycl file :P 

More rational, this allow automatic regeneration of the object file, and if one file needs regeneration then the other needs it as well.

cmake/Modules/FindComputeCpp.cmake
Naghasan: This output the kernel in the requested format, so it will be bitcode for SPIR or the binary blob for SPIRV etc. The .o is just used as a generic extension.

DuncanMcBain: Ah, I see. I was slightly confused by the .o suffix, so could we maybe switch to .bc? Or we could also use the value of COMPUTECPP_BITCODE, which would at least indicate what the file contains, even if it's not the "real" file extension. Let me know your thoughts :)

cmake/Modules/FindComputeCpp.cmake
mmha: It is not consumed by the SYCL program. Therefore it does not depend on it.

There's nothing really *wrong* here, it just doesn't feel clean.

Side note: you could make the object file a `BYPRODUCT` instead of an `OUTPUT` if it only serves debug purposes.

> More rational, this allow automatic regeneration of the object file, and if one file needs regeneration then the other needs it as well.

With this change the SYCL translation unit is rebuilt when the object file or the integration header changes. Since integration header and object file are built by the same command, there will be no semantic difference.

cmake/Modules/FindComputeCpp.cmake
Naghasan: No matter what we pick it is going to be confusing, it should be .s for ptx, .bc for SPIR and .spv for SPIRV...

DuncanMcBain: I was thinking about this and you could do something like:
```CMake
set(IR_MAP_spir64 bc)
set(IR_MAP_ptx64 s)
# later...
set(outputDeviceFile ${baseSyclName}.${IR_MAP_${COMPUTECPP_BITCODE}})
```

Kind of nasty, but would get accurate file names every time!

Naghasan: That looks nasty to me, but if this can avoid confusion why not

DuncanMcBain: I tried my hand at adding it, and I think putting it in its own file is the easiest option for now - I prepared a patch if that helps:
```diff
diff --git a/cmake/Modules/ComputeCppIRMap.cmake b/cmake/Modules/ComputeCppIRMap.cmake
new file mode 100644
index 0000000..b3dd770
--- /dev/null
+++ b/cmake/Modules/ComputeCppIRMap.cmake
@@ -0,0 +1,17 @@
+cmake_minimum_required(VERSION 3.4.3)
+
+# These should match the types of IR output by compute++
+set(IR_MAP_spir bc)
+set(IR_MAP_spir64 bc)
+set(IR_MAP_spir32 bc)
+set(IR_MAP_spirv spv)
+set(IR_MAP_spirv64 spv)
+set(IR_MAP_spirv32 spv)
+set(IR_MAP_aorta-x86_64 o)
+set(IR_MAP_aorta-aarch64 o)
+set(IR_MAP_custom-spir64 bc)
+set(IR_MAP_custom-spir32 bc)
+set(IR_MAP_custom-spirv64 spv)
+set(IR_MAP_custom-spirv32 spv)
+set(IR_MAP_ptx64 s)
+set(IR_MAP_amdgcn s)
diff --git a/cmake/Modules/FindComputeCpp.cmake b/cmake/Modules/FindComputeCpp.cmake
index 66d6538..98809bb 100644
--- a/cmake/Modules/FindComputeCpp.cmake
+++ b/cmake/Modules/FindComputeCpp.cmake
@@ -31,6 +31,7 @@
 
 cmake_minimum_required(VERSION 3.4.3)
 include(FindPackageHandleStandardArgs)
+include(ComputeCppIRMap)
 
 set(COMPUTECPP_USER_FLAGS "" CACHE STRING "User flags for compute++")
 separate_arguments(COMPUTECPP_USER_FLAGS)
@@ -230,7 +231,7 @@ function(__build_ir)
   # using the same source file will be generated with a different rule.
   set(baseSyclName ${CMAKE_CURRENT_BINARY_DIR}/${SDK_BUILD_IR_TARGET}_${sourceFileName})
   set(outputSyclFile ${baseSyclName}.sycl)
-  set(outputDeviceFile ${baseSyclName}.o)
+  set(outputDeviceFile ${baseSyclName}.${IR_MAP_${COMPUTECPP_BITCODE}})
   set(depFileName ${baseSyclName}.sycl.d)
 
   set(include_directories "$<TARGET_PROPERTY:${SDK_BUILD_IR_TARGET},INCLUDE_DIRECTORIES>")
```
I took the list of IR formats from `compute++` listing what the acceptable options to `-sycl-target` are.

PR #183, opened 2019-02-19
Fix non forward declarable SYCL kernel names

Naghasan:
Some kernel names were not forward declarable, which is illegal in SYCL 1.2.1.
Add a forward declaration to those kernel name to make them globally visible.

### DIFF
--- a/samples/opencl-c-interop/opencl-c-interop.cpp
+++ b/samples/opencl-c-interop/opencl-c-interop.cpp
@@ -33,6 +33,8 @@
 
 using namespace cl::sycl;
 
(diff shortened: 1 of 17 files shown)


samples/opencl-c-interop/opencl-c-interop.cpp
```diff
@@ -33,6 +33,8 @@
 
 using namespace cl::sycl;
 
+class pow_comp;
```
DuncanMcBain: This one hasn't removed the `class` from the call site, not an error but would be consistent

PR #184, opened 2019-02-21
Remove deprecated function calls from placeholder accessor test

ProGTX:
1. Re-assign accessors instead of calling `require`
2. Removed the use of `this` as a kernel


Ruyk:
@ProGTX  can you also remove the usage of c++14 lambdas from the test before merging?

PR #198, opened 2019-05-28
Adding Monte-Carlo Pi sample

GeorgeWeb:
Adding a new sample: **Monte-Carlo Pi** approximation algorithm, that makes use of `info::kernel_work_group::work_group_size` & `info::device::max_work_group_size` 
to query if the kernel can be executed.

### DIFF
--- a/samples/monte-carlo-pi/monte-carlo-pi.cpp
+++ b/samples/monte-carlo-pi/monte-carlo-pi.cpp
@@ -107,7 +107,7 @@ size_t get_best_work_group_size(size_t work_group_size,
                                 const sycl::kernel& kernel) {
   if (device.is_host()) {
     /* Maximum Work-Group Size on selected device.
-     * (See: sycl-1.2.1.pdf: p.180, table 4.85) */
+     * (See: sycl-1.2.1.pdf: p.62, table 4.20) */
     const size_t max_device_work_group_size =
         device.get_info<cl::sycl::info::device::max_work_group_size>();
     /* Check if the desired work-group size will be allowed on the host device
--- a/samples/monte-carlo-pi/monte-carlo-pi.cpp
+++ b/samples/monte-carlo-pi/monte-carlo-pi.cpp
@@ -1,0 +1,229 @@
+/***************************************************************************
+ *
+ *  Copyright (C) 2019 Codeplay Software Limited
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  For your convenience, a copy of the License has been included in this
+ *  repository.
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *  Codeplay's ComputeCpp SDK
+ *
+ *  monte-carlo-pi.cpp
+ *
+ *  Description:
+ *    Example of Monte-Carlo Pi approxamtion algorithm in SYCL.
+ *    Also demonstrating how to query the maximum number of work-items in a
+ *    work-group to check if a kernel can be executed with the initially set
+ *    size.
+ *
+ *
+ **************************************************************************/
+
+#include <CL/sycl.hpp>
+
+// Standard C++ includes
+#include <iostream>
+#include <random>
+#include <typeinfo>
+#include <vector>
+
+namespace sycl = cl::sycl;
+
+// Monte-Carlo Pi SYCL C++ functor
+class monte_carlo_pi_kernel {
+  template <typename dataT>
+  using read_const_accessor =
+      sycl::accessor<dataT, 1, sycl::access::mode::read,
+                     sycl::access::target::constant_buffer>;
+  template <typename dataT>
+  using write_global_accessor =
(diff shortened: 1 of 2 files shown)


samples/monte-carlo-pi/monte-carlo-pi.cpp
```diff
@@ -0,0 +1,255 @@
...
+  std::vector<sycl::cl_int> results;
+
+  // Generate random points on the host - one point for each thread
+  std::vector<sycl::float2> points;
+  points.reserve(iterations);
+  // fill up with random values
+  std::random_device r;
+  std::default_random_engine e(r());
+  std::uniform_real_distribution<float> dist;
+  for (size_t i = 0; i < iterations; i++) {
```
DuncanMcBain: It's not necessary, but you could use `std::generate` here.

samples/monte-carlo-pi/monte-carlo-pi.cpp
```diff
@@ -0,0 +1,255 @@
...
+                                sycl::access::target::constant_buffer>(cgh);
+      auto results_ptr = results_buf.get_access<sycl::access::mode::write>(cgh);
+      // Allocate local memory on the device (to compute results)
+      sycl::accessor<sycl::cl_int, 1, sycl::access::mode::write,
+                     sycl::access::target::local>
+          local_results_ptr(sycl::range<1>(local_size), cgh);
+
+      // Run the kernel
+      cgh.parallel_for(
+          program.get_kernel<monte_carlo_pi_kernel>(),
```
DuncanMcBain: `kernel` is still in scope here

samples/monte-carlo-pi/monte-carlo-pi.cpp
```diff
@@ -0,0 +1,255 @@
...
+
+  // Sum the results (auto copied back to host)
+  int in_circle = 0;
+  for (auto& result : results) {
+    in_circle += result;
+  }
+
+  // Calculate the final result of "pi"
+  float pi =
+      (4.0f * static_cast<float>(in_circle)) / static_cast<float>(iterations);
```
DuncanMcBain: You should be able to remove the static casts here since at least one of the operands is a float.

samples/monte-carlo-pi/monte-carlo-pi.cpp
```diff
@@ -0,0 +1,255 @@
...
+                << max_kernel_work_group_size << "\n\n";
+      return max_kernel_work_group_size;
+    }
+    // otherwise, the work-size will stay the originally desired one
+    return work_group_size;
+  }
+}
+
+int main() {
+  /* 2 ^ 20 = 262144, large enough and is not going to cause memory allocation
```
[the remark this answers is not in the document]
GeorgeWeb: This makes sense. In that case, should I remove both comments describing the chosen sizes (```iterations``` and ```work_group_size``` - desired one) ?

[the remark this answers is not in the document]
DuncanMcBain: With a good enough name and size, the comment is really only rephrasing what the code says - so I'd say it's fine to remove the comments, as that's just something else to keep in sync.

samples/monte-carlo-pi/monte-carlo-pi.cpp
```diff
@@ -0,0 +1,255 @@
...
+  template <typename dataT>
+  using read_const_accessor =
+      sycl::accessor<dataT, 1, sycl::access::mode::read,
+                     sycl::access::target::constant_buffer>;
+  template <typename dataT>
+  using write_global_accessor =
+      sycl::accessor<dataT, 1, sycl::access::mode::write,
+                     sycl::access::target::global_buffer>;
+  template <typename dataT>
+  using write_local_accessor =
```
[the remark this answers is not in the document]
ProGTX: Local accessors can only have read-write access, ComputeCpp will soon fail to compile with anything else.

samples/monte-carlo-pi/monte-carlo-pi.cpp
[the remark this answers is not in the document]
GeorgeWeb: Hmm, would `iterations = 1024 * 1024` and `work_group_size = 1024` be a viable change to make the current size definitions more obvious to readers?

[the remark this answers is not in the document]
DuncanMcBain: If you like. 1<<10 and 1 << 20 are also clear. I was just noting that the comment was slightly misleading.

samples/monte-carlo-pi/monte-carlo-pi.cpp
```diff
@@ -0,0 +1,248 @@
...
+  constexpr size_t iterations = 1 << 20;
+  size_t work_group_size = 1 << 10;
+
+  // Container for the sum calculated per each work-group.
+  std::vector<sycl::cl_int> results;
+
+  // Generate random points on the host - one point for each thread
+  std::vector<sycl::float2> points(iterations);
+  // Fill up with (pseudo) random values in the range: [0, 1]
+  std::generate(points.begin(), points.end(), []() {
```
DuncanMcBain: Since this is a lambda, I think it is better to create the random device and distribution etc. outside of the generate call and have them captured by reference inside the lambda.

GeorgeWeb: Definitely better! My initial idea was to have it defined `static` inside so it is slightly more compact, I mean same lines of code but seem more as a part of the lambda but I see how this was not a really good idea. 

Actually, if I have enclosed the whole random generation operation in a scope `{}` would've still done it for me but I guess I didn't think about it as much at the time. 

Anyway, I will just do as suggested. No need to debate on style or this will never get merged lol.

DuncanMcBain: Yeah the way you have it now is perfect!

PR #200, opened 2019-06-06
Update build guide for demos

GeorgeWeb:
Updates the build guide with a fix for building **Nbody** due to outdated dependency issue in `Cinder-ImGui`.

### DIFF
--- a/demos/README.md
+++ b/demos/README.md
@@ -32,11 +32,14 @@ cmake -DCMAKE_BUILD_TYPE=Release -DCINDER_BOOST_USE_SYSTEM=1 ..
 make
 ```
 
-Then navigate to the `Cinder-ImGui` folder where you have to change the commit of the `imgui` dependency. \
-After `cd lib/imgui`, checkout the working commit as follows:
+Then navigate to the Cinder-ImGui folder where you have to change the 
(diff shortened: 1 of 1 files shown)


demos/README.md
```diff
@@ -32,6 +32,13 @@ cmake -DCMAKE_BUILD_TYPE=Release -DCINDER_BOOST_USE_SYSTEM=1 ..
 make
 ```
 
+Then navigate to the `Cinder-ImGui` folder where you have to change the commit of the `imgui` dependency. \
```
DuncanMcBain: Could you keep lines to < 80 characters please?

demos/README.md
```diff
@@ -32,6 +32,13 @@ cmake -DCMAKE_BUILD_TYPE=Release -DCINDER_BOOST_USE_SYSTEM=1 ..
...
+After `cd lib/imgui`, checkout the working commit as follows:
+
+```
+git checkout df8a9c49eb6d9f134411eeffa0441f561aec3967
```
DuncanMcBain: You can just name the first 7 characters which is a little more readable. We should also include a reason why we need this (imgui has made changes, cinder-imgui doesn't seem to have caught up yet).

PR #213, opened 2019-10-02
Update vector-add example

codeplay-kostasrim:
+ Added idiomatic C++
+ Removed all raw for loops and replaced them with standard algorithms
+ Removed the different return codes (they were used wrong)
+ Renamed single letter containers
+ Refactored the code for readability


DuncanMcBain:
This is beyond the scope of the samples for the moment.

PR #214, opened 2019-10-02
Update accessor

codeplay-kostasrim:
- Removed magic constants
- Removed multiple return values from main


samples/accessors.cpp
```diff
@@ -27,17 +27,15 @@
...
-using namespace cl::sycl;
-
 class multiply;
 
 int main() {
+  using namespace cl::sycl;  
   /* We define the data to be passed to the device. */
   int data = 5;
-
+
```
DuncanMcBain: There's a clang-format-6 compatible config file in the SDK, would appreciate if you could run it over the files you touch

samples/accessors.cpp
```diff
@@ -75,15 +73,14 @@ int main() {
 
   } catch (exception const& e) {
     std::cout << "SYCL exception caught: " << e.what() << '\n';
-    return 2;
+    return 0;
```
DuncanMcBain: Appreciate the discussion around this, but to keep the test failing when there is some error it would be good if the sample could rethrow the exception

samples/accessors.cpp
```diff
@@ -75,15 +73,14 @@ int main() {
...
   }
 
+  constexpr int expectedResult = 5 * 5;  
   /* We check that the result is correct. */
-  if (data == 25) {
-    std::cout << "Hurray! 5 * 5 is " << data << '\n';
-    return 0;
-  } else {
-    std::cout << "Oops! Something went wrong... 5 * 5 is not " << data << "!\n";
-    return 1;
```
DuncanMcBain: I don't want this removed because I want incorrect data to be a failure such that ctest marks it as failed (unlikely, I know, but still).

PR #216, opened 2019-10-02
Fix the scan example to avoid undefined behavior.

jcranmer-intel:
To prevent the original input data from being overwritten, the buffer
constructor with the const pointer input needs to be used. When the
non-const pointer variant is used, the implementation is free to
overwrite the original buffer pointer (if, for example, it chose to use
the CL_MEM_HOST_PTR flag on the underlying OpenCL call), even if it is
being written back to a different data pointer.

### DIFF
--- a/samples/scan.cpp
+++ b/samples/scan.cpp
@@ -257,7 +257,7 @@ int test_sum(sycl::queue& q) {
   std::vector<int32_t> sum(in.size());
   {
     // Read from `in`, but write into `sum`.
-    sycl::buffer<int32_t, 1> buf(in.data(), sycl::range<1>(in.size()));
+    sycl::buffer<int32_t, 1> buf((const int32_t*)in.data(), sycl::range<1>(in.size()));
     buf.set_final_data(sum.data());
 
     par_scan<int32_t, std::plus<int32_t>>(buf, q);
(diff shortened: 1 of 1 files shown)


DuncanMcBain:
Hi @jcranmer-intel,

Thanks for the fix! You're right, this could lead to unintended behaviour under the right circumstances. However, rather than a cast, we'd prefer something like the following as a fix:
```udiff
diff --git a/samples/scan.cpp b/samples/scan.cpp
index 42d0ba0..194e9f9 100644
--- a/samples/scan.cpp
+++ b/samples/scan.cpp
@@ -257,8 +257,12 @@ int test_sum(sycl::queue& q) {
   std::vector<int32_t> sum(in.size());
   {
     // Read from `in`, but write into `sum`.
-    sycl::buffer<int32_t, 1> buf(in.data(), sycl::range<1>(in.size()));
+    sycl::buffer<int32_t, 1> buf(sycl::range<1>(in.size()));
     buf.set_final_data(sum.data());
+    q.submit([&](sycl::handler& cgh) {
+      auto acc = buf.get_access<sycl::access::mode::write>(cgh);
+      cgh.copy(in.data(), acc);
+    });
 
     par_scan<int32_t, std::plus<int32_t>>(buf, q);
   }
@@ -299,8 +303,12 @@ int test_factorial(sycl::queue& q) {
   std::vector<int64_t> fact(in.size());
   {
     // Read from `in`, but write into `fact`.
-    sycl::buffer<int64_t, 1> buf(in.data(), sycl::range<1>(in.size()));
+    sycl::buffer<int64_t, 1> buf(sycl::range<1>(in.size()));
     buf.set_final_data(fact.data());
+    q.submit([&](sycl::handler& cgh) {
+      auto acc = buf.get_access<sycl::access::mode::write>(cgh);
+      cgh.copy(in.data(), acc);
+    });
 
     par_scan<int64_t, std::multiplies<int64_t>>(buf, q);
   }
```
This makes the intent more clear (i.e. that the buffer is to be used on the device only). I didn't want to ask you to write a larger change on our behalf, since it was our mistake in the first place!

Thanks,
Duncan.

DuncanMcBain:
I pushed our patch in 22c32351b9f82c4edcc475797e2e6e6560081a9e. Thanks for making us aware of this problem!

PR #219, opened 2019-11-29
Update SDK to ComputeCpp v1.1.6

DuncanMcBain:
This version of ComputeCpp introduced the index flip change. This only
affects code that does its own index calculations, using the id()
functions to index will continue to work.

### DIFF
--- a/samples/template-function-object.cpp
+++ b/samples/template-function-object.cpp
@@ -60,7 +60,7 @@ class vector_add_kernel {
    * item rather than an nd_item. */
   void operator()(item<1> item) {
     /* This is a standard vector add, as seen in other samples. */
-    m_ptrC[item] = m_ptrA[item] + m_ptrB[item];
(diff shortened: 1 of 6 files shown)


samples/template-function-object.cpp
```diff
@@ -60,7 +60,7 @@ class vector_add_kernel {
    * item rather than an nd_item. */
   void operator()(item<1> item) {
     /* This is a standard vector add, as seen in other samples. */
-    m_ptrC[item.get_id()] = m_ptrA[item.get_id()] + m_ptrB[item.get_id()];
+    m_ptrC[item] = m_ptrA[item] + m_ptrB[item];
```
bader: Buffer accessor subscript operator has no overloads with `item` parameter. This changes is not valid according to the SYCL specification.
The same is true for the change in `samples/using-function-objects.cpp`.

DuncanMcBain: You are quite correct, it's a quirk of our implementation. I've removed the changes made in this PR and have made an issue to check in case it has sneaked its way into any other code. Thanks!

PR #231, opened 2020-04-09
Add missing header file <map> to virtual pointer sample

Srividya2212:
The vptr.hpp file used in the virtual pointer sample was using std::map but was missing <map> the header file <map>
I have added the <map> header to vptr.hpp to fix the compilation errors.

### DIFF
--- a/include/vptr/virtual_ptr.hpp
+++ b/include/vptr/virtual_ptr.hpp
@@ -39,7 +39,6 @@
 #include <queue>
 #include <set>
 #include <stdexcept>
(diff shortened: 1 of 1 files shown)


Srividya2212:
@Ruyk I have removed <unordered_map> as well, since it wasn't used.
How can I merge this PR. I don't think I have write access :
`Only those with write access to this repository can merge pull requests.`

Ruyk:
I'll do it, thanks for your contribution!

PR #234, opened 2020-04-29
Update SDK to ComputeCpp v2.0.0

DuncanMcBain:
The name of the ComputeCpp library has changed on Windows. Since
we are supporting MSVC 2019 now, we need to include the compiler
checks to make it work, and also need an update to Google Test
to avoid a deprecation in MSVC.

### DIFF
--- a/cmake/Modules/FindComputeCpp.cmake
+++ b/cmake/Modules/FindComputeCpp.cmake
@@ -74,13 +74,13 @@ find_program(ComputeCpp_INFO_EXECUTABLE computecpp_info
   NO_SYSTEM_ENVIRONMENT_PATH)
 
 find_library(COMPUTECPP_RUNTIME_LIBRARY
-  NAMES ComputeCpp ComputeCpp_vs2015
+  NAMES ComputeCpp
   HINTS ${computecpp_find_hint}
   PATH_SUFFIXES lib
   DOC "ComputeCpp Runtime Library")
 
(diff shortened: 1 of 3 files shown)


cmake/Modules/FindComputeCpp.cmake
```diff
@@ -399,6 +403,8 @@ function(add_sycl_to_target)
     ${ARGN}
   )
 
+  set_target_properties(${SDK_ADD_SYCL_TARGET} PROPERTIES LINKER_LANGUAGE CXX)
```
jwlawson, line 406: Nit: Obscure changes like this should probably have some explanation in the commit.

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -74,13 +74,13 @@ find_program(ComputeCpp_INFO_EXECUTABLE computecpp_info
   NO_SYSTEM_ENVIRONMENT_PATH)
 
 find_library(COMPUTECPP_RUNTIME_LIBRARY
-  NAMES ComputeCpp ComputeCpp_vs2015
```
jwlawson, line 77: Nit: I'd be tempted to leave the old library names available, so this doesn't just break when using the old windows packages.

DuncanMcBain:
We do squashes on the SDK and there are unresolved comments. Please don't merge when there are unresolved comments

PR #236, opened 2020-05-03
RelWithDebInfo: Has to be compiled with release options

ThijsWithaar:
In de cmake script, RELWITHDEBINFO uses the debug libraries.
Although these do have debug symbols, they are not compatible with a release build.

In order to link succesfully, the release libs should be used instead.


DuncanMcBain:
Thanks for this fix!

PR #241, opened 2020-08-19
Fix missing header in sync-handler sample

alexbatashev:
`std::any_of` requires `#include <algorithm>`.

Signed-off-by: Alexander Batashev <alexander.batashev@intel.com>


DuncanMcBain (approved):

DuncanMcBain:
Thanks for the fix!

PR #242, opened 2020-08-25
Make CMake fail when compute++ can't compile code

mcleary:
- compute++ 2.1.0 is based on clang 8 and it can't parse the Windows headers for MSVC 19.27.
- This patch will make CMake try to build a simple program that includes cl/SYCL.hpp
- If the compilation fails we try to compile with __CUDACC__ and HAS_CONDITIONAL_EXPLICIT=0. If compute++ can build the file, those flags are added to COMPUTECPP_DEVICE_COMPILER_FLAGS

### DIFF
--- a/cmake/Modules/ComputeCppCompilerChecks.cmake
+++ b/cmake/Modules/ComputeCppCompilerChecks.cmake
@@ -13,16 +13,17 @@ endif()
 if(MSVC)
   set(ComputeCpp_STL_CHECK_SRC __STL_check)
   file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/${ComputeCpp_STL_CHECK_SRC}.cpp
-    "#include <CL/sycl.hpp>\n"
+    "#include <CL/sycl.hpp>  \n"
     "int main() { return 0; }\n")
+  set(_stl_test_command ${ComputeCpp_DEVICE_COMPILER_EXECUTABLE}
+                        -sycl
+                        ${COMPUTECPP_DEVICE_COMPILER_FLAGS}
+                        -isystem ${ComputeCpp_INCLUDE_DIRS}
+                        -isystem ${OpenCL_INCLUDE_DIRS}
+                        -o ${ComputeCpp_STL_CHECK_SRC}.sycl
+                        -c ${ComputeCpp_STL_CHECK_SRC}.cpp)
   execute_process(
-    COMMAND ${ComputeCpp_DEVICE_COMPILER_EXECUTABLE}
-            -sycl
-            ${COMPUTECPP_DEVICE_COMPILER_FLAGS}
-            -isystem ${ComputeCpp_INCLUDE_DIRS}
-            -isystem ${OpenCL_INCLUDE_DIRS}
-            -o ${ComputeCpp_STL_CHECK_SRC}.sycl
-            -c ${ComputeCpp_STL_CHECK_SRC}.cpp
+    COMMAND ${_stl_test_command}
     WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
     RESULT_VARIABLE ComputeCpp_STL_CHECK_RESULT
     ERROR_QUIET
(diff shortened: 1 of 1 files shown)


cmake/Modules/ComputeCppCompilerChecks.cmake
```diff
@@ -38,10 +38,16 @@ if(MSVC)
       RESULT_VARIABLE ComputeCpp_STL_CHECK_RESULT
       ERROR_QUIET
       OUTPUT_QUIET)
-    if(NOT ${ComputeCpp_STL_CHECK_RESULT} EQUAL 0)
-      message(STATUS "Device compiler cannot consume hosted STL headers. Using any parts of the STL will likely result in device compiler errors.")
-    else()
-    message(STATUS "Device compiler does not meet certain STL version requirements. Disabling version checks and hoping for the best.")
+    if(NOT ${ComputeCpp_STL_CHECK_RESULT} EQUAL 0)      
+      message(STATUS "Device compiler cannot consume hosted STL headers. Using any parts of the STL will likely result in device compiler errors")
+      if (${MSVC_VERSION} GREATER_EQUAL 1927)
```
DuncanMcBain, line 43: I'm really not a fan of putting hardcoded error checks like this in CMake. It ties this to a specific (but potentially unknown) version of ComputeCpp, and this will hang around. People might copy it and therefore it won't get updated. Or, as I worry, we'll fix this and forget to update the check here.

The best way would be to force the failure in the testing code we're already running. Do you know which header causes the compilation failure? We'd then be able to add it at line 16, and actually not really need any other changes.

Maybe we should be adding all the headers that ComputeCpp includes in its own headers to the test file, since that is completely representative.

DuncanMcBain (changes requested):
It's good that we're updating this but I think we should just make sure that the original mechanism continues to identify our failure cases correctly

cmake/Modules/ComputeCppCompilerChecks.cmake
mcleary: > Do you know which header causes the compilation failure? We'd then be able to add it at line 16, and actually not really need any other changes.

It's not that simple. On Windows, the header changes depending on the selected toolset, that's why the failure is happening now. The xutility header is fine in version 14.26 but the device compiler can't parse xutility in 14.27.

I understand your concern and I don't like to have these hard-coded numbers either, but this was the only way I could think to warn users about this problem.

DuncanMcBain: What I mean is, if you add the headers that are transitively included by Codeplay's `CL/sycl.hpp` to the test file, it will in turn include `<xutility>`, and if that header is from the troublesome version then we'll get a failure as expected. We could maybe have a better message (perhaps pointing users to the support forum, or similar?) but the general idea in this file is "if this breaks your Visual Studio is too new for us".

mcleary: Ah right, in this case we don't need to do anything. The existing message will be showed, but just as a warning. I changed it to a fatal error because I almost didn't notice it. I will try to improve the message then

DuncanMcBain: So if we do add the std headers from `CL/sycl.hpp` and that compile fails, it's fine to mark it as a `FATAL_ERROR`, because literally just including the header will fail the build. If it requires the define we can downgrade it to a `WARNING`, with the caveat of it might still fail depending on what gets used.

mcleary: That's the thing, you don't need to include anything from sycl, just by using the device compiler to build this

```c++
#include <ios>
int main() {return 0;}
```

Will fail. I will simply update the message to a FATAL_ERROR then

DuncanMcBain: I see what you mean. Would you be able to update the list then fatal error if it can't compile? The reason I ask is I know we don't include `<ios>` in ComputeCpp user headers so it gives us a more accurate idea of whether it will work or not. If we're forcibly including a header which will always fail the compile in an otherwise empty file then I'm more happy with the FATAL_ERROR. Does that make sense?

cmake/Modules/ComputeCppCompilerChecks.cmake
```diff
@@ -38,10 +38,16 @@ if(MSVC)
       RESULT_VARIABLE ComputeCpp_STL_CHECK_RESULT
       ERROR_QUIET
       OUTPUT_QUIET)
-    if(NOT ${ComputeCpp_STL_CHECK_RESULT} EQUAL 0)
-      message(STATUS "Device compiler cannot consume hosted STL headers. Using any parts of the STL will likely result in device compiler errors.")
-    else()
-    message(STATUS "Device compiler does not meet certain STL version requirements. Disabling version checks and hoping for the best.")
+    if(NOT ${ComputeCpp_STL_CHECK_RESULT} EQUAL 0)
```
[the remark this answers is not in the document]
mcleary: fixed

cmake/Modules/ComputeCppCompilerChecks.cmake
```diff
@@ -38,10 +38,16 @@ if(MSVC)
       RESULT_VARIABLE ComputeCpp_STL_CHECK_RESULT
       ERROR_QUIET
       OUTPUT_QUIET)
-    if(NOT ${ComputeCpp_STL_CHECK_RESULT} EQUAL 0)
-      message(STATUS "Device compiler cannot consume hosted STL headers. Using any parts of the STL will likely result in device compiler errors.")
-    else()
-    message(STATUS "Device compiler does not meet certain STL version requirements. Disabling version checks and hoping for the best.")
+    if(NOT ${ComputeCpp_STL_CHECK_RESULT} EQUAL 0)      
+      message(STATUS "Device compiler cannot consume hosted STL headers. Using any parts of the STL will likely result in device compiler errors")
```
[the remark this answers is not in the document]
mcleary: Fixed

cmake/Modules/ComputeCppCompilerChecks.cmake
mcleary: I've updated message to a fatal error and added some text explaining a possible solution for the problem

mcleary:
It turns out that the MSVC headers use a less restricted version of some functions when __CUDACC__ is defined and the device compiler can parse the STL headers in the latest update (14.27). I've added a check to add this define to COMPUTECPP_DEVICE_COMPILER_FLAGS, which allows the device compiler to be used with the latest VS update.

DuncanMcBain (approved):
This is a great change! Since this is now so much more robust, maybe we could auto-include this file in FindComputeCpp now? A lot of our windows users are going to experience this now and this has gone from an occasional annoyance to majority of folk will see it.

It also might be nice to factor the `COMMAND` part of the `execute_process` into a variable to save some lines, but I'm not going to block the PR on that (it's not that important).

mcleary:
> This is a great change! Since this is now so much more robust, maybe we could auto-include this file in FindComputeCpp now? A lot of our windows users are going to experience this now and this has gone from an occasional annoyance to majority of folk will see it.

The file is already included in FindComputeCpp

> 
> It also might be nice to factor the `COMMAND` part of the `execute_process` into a variable to save some lines, but I'm not going to block the PR on that (it's not that important).

mcleary:
> It also might be nice to factor the `COMMAND` part of the `execute_process` into a variable to save some lines, but I'm not going to block the PR on that (it's not that important).

Done

DuncanMcBain (approved):

DuncanMcBain:
You're right, someone updated it four months ago. :thinking:

PR #248, opened 2020-10-16
Allow non-standard template parameters in accessor

alexbatashev:
This only works if additional template parameters have default value,
and that value is used when passing type to is_accessor trait.


alexbatashev:
Tagging @romanovvlad

PR #249, opened 2020-11-30
SYCL_LANGUAGE_VERSION option

ProGTX:
ComputeCpp 2.2.1 provides support for SYCL 2020.
This patch exposes that by providing the `SYCL_LANGUAGE_VERSION` option.
Set for host and device compilers. Defaults to SYCL 1.2.1.

### DIFF
--- a/cmake/Modules/FindComputeCpp.cmake
+++ b/cmake/Modules/FindComputeCpp.cmake
@@ -41,6 +41,8 @@ set(COMPUTECPP_BITCODE "spir64" CACHE STRING
   "Bitcode type to use as SYCL target in compute++")
 mark_as_advanced(COMPUTECPP_BITCODE)
 
+set(SYCL_LANGUAGE_VERSION "2017" CACHE STRING "SYCL version to use. Defaults to 1.2.1.")
+
 find_package(OpenCL REQUIRED)
 
 # Find ComputeCpp package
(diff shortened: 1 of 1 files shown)


cmake/Modules/FindComputeCpp.cmake
```diff
@@ -41,6 +41,9 @@ set(COMPUTECPP_BITCODE "spir64" CACHE STRING
   "Bitcode type to use as SYCL target in compute++")
 mark_as_advanced(COMPUTECPP_BITCODE)
 
+set(SYCL_LANGUAGE_VERSION "2017" CACHE STRING "SYCL version to use. Defaults to 1.2.1.")
+mark_as_advanced(SYCL_LANGUAGE_VERSION)
```
DuncanMcBain, line 45: It's probably fine to leave this as shown to users. Folk might want to change it. We might be too eager to mark some of these variables as advanced, probably my mistake. Does 2017 == 1.2.1? That's a little unclear to me

DuncanMcBain (approved):

cmake/Modules/FindComputeCpp.cmake
ProGTX: `2017` is not an official value, just something we've decided to use since 1.2.1 came out in 2017. I expect this option to be mainly used to enable SYCL 2020 (and beyond) features, so the value for 1.2.1 shouldn't be problematic.

Will unmark it as advanced.

rodburns (approved):

mcleary (approved):

PR #250, opened 2021-01-07
Support for multiple bitcode targets in COMPUTECPP_BITCODE

mcleary:
Adds support for setting multiple values in `COMPUTECPP_BITCODE`. When using ComputeCpp Professional Edition this allows multiple versions of the same kernel to be available in the same binary.


ProGTX (approved):
I've made some changes, mostly fixing the filename extension mapping and reverting `spirv64` because that has some larger implications.

PR #252, opened 2021-03-04
Switch from Travis to Github Actions

Bensuo:
Moves CI testing from Travis to Github Actions, given the impending shutdown of Travis' `.org` service.

Changes:

- Adds a Github Actions workflow which replicates the original Travis workflow. Runs on pushes to `master` as well as pull requests and on-demand through the Actions tab.
- Adds a `Dependabot` config to track updates to Actions dependancies.
- Removes the `.travis.yml` file.
- Fixed missing SDK logo in `README.md` and replaced Travis status badge with Actions badge.


README.md
```diff
@@ -1,11 +1,12 @@
-# ComputeCpp SDK Readme
 
-![ComputeCpp™]
+<div align="center">
```
DuncanMcBain, line 2: Is there no way to link this new image in the old style? If you check the bottom of the document, that's where we've defined the links to keep the main body a bit more readable. At the very least, the old links should be removed from there, though I'd prefer that we add the links there instead.

README.md
```diff
@@ -1,11 +1,12 @@
...
+  <br /><img src="https://developer.codeplay.com/assets/img/home/tools/computecppsdk.png"><br />
+</div>
 
+# ComputeCpp SDK Readme
+[![Build and Test](https://github.com/codeplaysoftware/computecpp-sdk/actions/workflows/build-test.yml/badge.svg)](https://github.com/codeplaysoftware/computecpp-sdk/actions/workflows/build-test.yml)
```
DuncanMcBain, line 7: Similarly, I'd hope that we can update the travis SVG link at the end of the file to the new shiny actions one you have here, which keeps things a bit more organised (while still leaving a useful link inline).

.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: daily
```
DuncanMcBain, line 6: Will this build the SDK once per day? If so, I'd not bother. Provided it builds master when it's updated, that should be fine. (What do we gain from daily builds?)

DuncanMcBain (changes requested):
I can't leave a comment on it, but I'd probably keep the `build_computecpp.sh` file in the `.github` folder or similar. It's not used anywhere else, which is why it was in `.travis` before.

The actions stuff looks really good though!

.github/dependabot.yml
DuncanMcBain: I see an answer to this elsewhere!

README.md
Bensuo: Ah that is my mistake, I totally missed the links being defined at the end of the file! I've reverted this to the old style and updated the link for it at the bottom of the document.

README.md
Bensuo: I've also reverted this to the old style and updated the links!

Bensuo:
> I can't leave a comment on it, but I'd probably keep the `build_computecpp.sh` file in the `.github` folder or similar. It's not used anywhere else, which is why it was in `.travis` before.

The script has now been moved to the `.github` folder!

DuncanMcBain (approved):

DuncanMcBain:
Please use squash + rebase for PRs to this repo in future, thanks!

PR #254, opened 2021-03-30
[Build] Option to include integration header last

ProGTX:
Added `COMPUTECPP_SDK_INCLUDE_AFTER` option
that affects all SYCL targets by setting the `COMPUTECPP_INCLUDE_AFTER` property.

The option is disabled by default for now.

Additional changes:
* Make ComputeCpp_DIR a cache variable
* Updated .gitignore

### DIFF
--- a/cmake/Modules/FindComputeCpp.cmake
+++ b/cmake/Modules/FindComputeCpp.cmake
@@ -47,22 +47,19 @@ find_package(OpenCL REQUIRED)
 
 # Find ComputeCpp package
 
-if(DEFINED ComputeCpp_DIR)
-  set(computecpp_find_hint ${ComputeCpp_DIR})
-elseif(DEFINED ENV{COMPUTECPP_DIR})
-  set(computecpp_find_hint $ENV{COMPUTECPP_DIR})
-endif()
+set(ComputeCpp_DIR "$ENV{COMPUTECPP_DIR}"
+  CACHE PATH "Path to ComputeCpp root folder")
+# ComputeCpp_HOST_DIR is used to find executables that are run on the host
+set(ComputeCpp_HOST_DIR "$ENV{ComputeCpp_HOST_DIR}"
+  CACHE PATH "Path to host ComputeCpp root folder whe cross-compiling")
 
-# Used for running executables on the host
-set(computecpp_host_find_hint ${computecpp_find_hint})
+set(computecpp_find_hint "${ComputeCpp_DIR}")
 
-if(CMAKE_CROSSCOMPILING)
-  # ComputeCpp_HOST_DIR is used to find executables that are run on the host
-  if(DEFINED ComputeCpp_HOST_DIR)
-    set(computecpp_host_find_hint ${ComputeCpp_HOST_DIR})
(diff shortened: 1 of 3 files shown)


cmake/Modules/FindComputeCpp.cmake
```diff
@@ -47,22 +47,19 @@ find_package(OpenCL REQUIRED)
 
 # Find ComputeCpp package
 
-if(DEFINED ComputeCpp_DIR)
```
DuncanMcBain, line 50: This is a behaviour change and I don't understand it. (Apart from anything else, you're going to break my builds.)

What's the intent? I believe it also is counter to the CMake guidelines for how to write Find files.

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -328,6 +325,10 @@ function(__build_ir)
     add_custom_target(${headerTargetName} DEPENDS ${outputDeviceFile} ${outputSyclFile})
     add_dependencies(${SDK_BUILD_IR_TARGET} ${headerTargetName})
   endif()
+  
+  if(COMPUTECPP_SDK_INCLUDE_AFTER)
```
DuncanMcBain, line 329: It's a little strange to have the SDK naming if it's in the FindComputeCpp file. I also think it makes more sense to have it as part of `add_sycl_to_target`.

That being said, I also think it's a little weird to have a large behaviour change like this exposed via a configuration option (that's why I wrote it as a per-target property in the first place). I know there are reasons why we're considering changes like this, but don't we just want to change the default? Is the idea that if a project needs include after in any file, this is a simple way of making every target use it?

DuncanMcBain (changes requested):
I don't think the `ComputeCpp_DIR` related changes are correct (it should not be only an env var, or a CMake cache variable either).

cmake/Modules/FindComputeCpp.cmake
ProGTX: Why would this break your build? What's the use case?

The new behaviour should be the same as the old one, unless I'm missing something:
* If neither `ComputeCpp_DIR` nor `COMPUTECPP_DIR` env variable are defined, `ComputeCpp_DIR ` is set to empty, and the CMake configuration fails.
* The user can specify either variable, and it will be picked up by CMake. The CMake variable will have priority over the environment one.

Do CMake guidelines forbid options and cache variables in Find files? In that case we probably should drop this, yes.

cmake/Modules/FindComputeCpp.cmake
ProGTX: This seemed like the easiest way to achieve the desired behaviour, which yes is to be able to switch the entire project to use this property. I suppose `add_sycl_to_target` would be better, and I agree about the name. Is it possible to have an option that's named the same as a property? I don't think that should cause conflicts.

cmake/Modules/FindComputeCpp.cmake
DuncanMcBain: I see I'd slightly misread the second part of this change. Nevertheless AIUI the xxxx_DIR variable should not be cached so that the user can specify a different one in configuration.

PR #259, opened 2021-06-14
Fix Mandelbrot demo for hardware which doesn't support doubles

martinstarkov:
After messing with templates and trying to make the demo work with both floats and doubles depending on hardware, I noted that this overcomplicates the demo and makes it harder for beginners to focus on the point of the demo (which is to showcase a SYCL application). Floats work on all hardware hence changing doubles to floats makes the demo more general (without overcomplicating the source files).


DuncanMcBain:
The code you've changed here is entirely host-side code. Which hardware doesn't support `double` on the host CPU? Aside from that, I'd be pretty disappointed if it were so hard for us to have a demo which can run in both single- and double-precision modes. Surely the changes aren't that drastic?

martinstarkov:
> The code you've changed here is entirely host-side code. Which hardware doesn't support `double` on the host CPU? Aside from that, I'd be pretty disappointed if it were so hard for us to have a demo which can run in both single- and double-precision modes. Surely the changes aren't that drastic?

The code I changed modifies the device kernel as the class template takes in a float which will change the kernel too. My Intel(R) Core(TM) i5-5200U CPU drivers for OpenCL do not support double floating-point precision and I'm sure there are other examples. I think you are correct that perhaps the demo should support both but from my attempts it has been very difficult to get it to work since the source files need to be separated with the float/double instantiation but are instantiated already if you call them in an if-statement (checking if device supports double floating point precision). I'm not sure how to resolve this issue and I honestly feel like the demo will lose a lot of clarity if we have two source files. People in the SYCL-ECO chat were not able to help me when I was attempting a fix so I figured this would be simpler and works. The demo still runs well with floats and the user can zoom in for a very reasonable amount.

If you have any suggestions to support something like this (pseudo code) then please do let me know:

```c++
void calc() {
   if (device.get_info<???>()  == ???) {
      internal_calc<double>();
   } else {
      internal_calc<float>();
  }
}
```
I was not able to figure out how to separate the template instantiation into two sources files while having the above call be in one file..

martinstarkov:
Made the changes requested by @DuncanMcBain
I believe my approach to checking for double support is sufficient based on SYCL 2020 spec.

martinstarkov (commented):
The title of [631cfc3](https://github.com/codeplaysoftware/computecpp-sdk/pull/259/commits/631cfc3e31fe037b00e41286ee5c18239970c20c) commit should be "Change floats to doubles and check for SYCL device floating-point support" Got my words mixed up there.

demos/mandelbrot/mandel-float.cpp
```diff
@@ -0,0 +1,33 @@
...
+ *    Translation unit for the Mandelbrot SYCL kernel.
+ *
+ **************************************************************************/
+
+#include "mandel.hpp"
+
+template <>
+void MandelbrotCalculator::calc<float>() {
+   this->internal_calc<float>();
+}
```
DuncanMcBain, line 33: Can you run clang-format on these files? There's a couple of other things that look like they need formatted correctly, too. There's a clang-format file in the repo which has the right formatting options set.

DuncanMcBain (changes requested):
looks good!

demos/mandelbrot/mandel-float.cpp
```diff
@@ -0,0 +1,33 @@
...
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *  Codeplay's ComputeCpp SDK
+ *
+ *  mandel.cpp
```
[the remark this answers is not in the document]
martinstarkov: When you say 'remove' (which I'd be in favor of) do you mean the entire license in the file or just these 4 lines?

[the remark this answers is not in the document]
DuncanMcBain: The filenames, they're arguably not that useful. Not the whole preamble (that's a separate discussion for some other time). It's either that or update them to be accurate, I don't particularly mind which.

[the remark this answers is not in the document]
martinstarkov: Removed

demos/mandelbrot/mandel-float.cpp
martinstarkov: Done

martinstarkov:
Made all the requested changes

demos/mandelbrot/mandel-float.cpp
DuncanMcBain: Hmmm, the newlines are still missing at the end of the file. I thought that had been set in the clang-format file properly. Can you add them in?

DuncanMcBain: OK it appears that this is not a thing that `clang-format` will do for us, so for now can you add the single newline at the end of the file in?

martinstarkov: Added those

DuncanMcBain (approved):

PR #270, opened 2021-10-08
Update CMake version to version 3.10.2

DuncanMcBain:
This version comes with Ubuntu 18.04, and allows us to remove some
legacy OpenMP code. We are also better about dealing with some
policies that were introduced in newer CMake versions that are hit
by our code.


DuncanMcBain:
I would like to merge this because it is now blocking #283 as that sample needs to set a different CXX standard to the rest of the samples

Rbiessy (approved):
LGTM

Issue #95, opened 2018-02-28
constexpr problems with MSVC


ProGTX:
I've noticed that the Gaussian blur sample fails to compile with Visual Studio 2015 using ComputeCpp 0.6:
```
gaussian-blur.cpp(159): error C2131: expression did not evaluate to a constant
gaussian-blur.cpp(159): note: failure was caused by non-constant arguments or reference to a non-constant symbol
```
This is referring to this line of the `GaussianKernel` kernel lambda:
```
constexpr auto offset = 3 * stddev;
```
Even though `stddev` is a `constexpr` variable that was declared outside the kernel lambda, it's not considered `constexpr` inside the kernel lambda.

I believe this is a [Visual Studio bug](https://developercommunity.visualstudio.com/content/problem/93922/constexprs-are-trying-to-be-captured-in-lambda-fun.html) that's not even been fixed in MSVC 2017 (https://godbolt.org/g/t7n1uW).

Interestingly, it seems that making the `constexpr` variable `static constexpr` solves this problem, so maybe this should be the default way of using `constexpr` variables inside kernels.

DuncanMcBain:
Hi, I've pushed a commit that should fix it, though we don't have testing for Windows set up on GitHub. Could you confirm that this fix makes the Gaussian blur work on Windows?

DuncanMcBain:
Fixed in #98

PR #276, opened 2021-11-19
Fix usage of constexpr constants in MSVC

DuncanMcBain:
Seemingly MSVC cannot consume the constexpr constants in the
lambda without them being static as well.


hjabird (approved):
Looks good to me.

ProGTX:
Linking to https://github.com/codeplaysoftware/computecpp-sdk/issues/95 as related

PR #280, opened 2022-01-25
Update COMPUTECPP_BITCODE to let compiler decide SPIR/SPIR-V by default

DBDuncan:
Before, COMPUTECPP_BITCODE was set to a default of spir64. As we are aiming for SPIR-V to be the default this need to be changed. COMPUTECPP_BITCODE is now empty by default and when COMPUTECPP_BITCODE is empty, -sycl-target flag will not be used, letting the compiler decide whether to use SPIR or SPIR-V.

### DIFF
--- a/cmake/Modules/FindComputeCpp.cmake
+++ b/cmake/Modules/FindComputeCpp.cmake
@@ -63,7 +63,7 @@ set(COMPUTECPP_USER_FLAGS "" CACHE STRING "User flags for compute++")
 separate_arguments(COMPUTECPP_USER_FLAGS)
 mark_as_advanced(COMPUTECPP_USER_FLAGS)
 
-set(COMPUTECPP_BITCODE "spir64" CACHE STRING
+set(COMPUTECPP_BITCODE "" CACHE STRING
   "Bitcode types to use as SYCL targets in compute++.")
 mark_as_advanced(COMPUTECPP_BITCODE)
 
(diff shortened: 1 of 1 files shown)


ProGTX (approved):

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -63,7 +63,7 @@ set(COMPUTECPP_USER_FLAGS "" CACHE STRING "User flags for compute++")
 separate_arguments(COMPUTECPP_USER_FLAGS)
 mark_as_advanced(COMPUTECPP_USER_FLAGS)
 
-set(COMPUTECPP_BITCODE "spir64" CACHE STRING
+set(COMPUTECPP_BITCODE "" CACHE STRING
```
DuncanMcBain, line 66:
```suggestion
set(COMPUTECPP_BITCODE "" CACHE LIST
```

If you make this change, you can actually leave the code on line 188 onwards unchanged, since empty `foreach` loops don't actually execute the body at all.

DuncanMcBain (changes requested):

cmake/Modules/FindComputeCpp.cmake
DBDuncan: Yeah, that works. The only issue I guess is that it does spit out this warning.
```
CMake Warning (dev) at cmake/Modules/FindComputeCpp.cmake:66 (set):
  implicitly converting 'LIST' to 'STRING' type.
```

DBDuncan: Going to leave it out. Would rather avoid the warning.

Bensuo (approved):

cmake/Modules/FindComputeCpp.cmake
ProGTX: CACHE variables cannot be LIST: https://cmake.org/cmake/help/latest/command/set.html#set-cache-entry

PR #286, opened 2022-09-15
Change virtual pointer to use vptr namespace instead of cl::sycl::codeplay namespace

hjabird:
The vptr class is used in SYCL-BLAS. We want to be able to compile SYCL-BLAS using DPC++. The latest DPC++ nightlies generates errors due to the cl::sycl namespace used for virtual pointers. To avoid this issue, this patch uses the SYCL-2020 ::sycl namespace for DPC++.

### DIFF
--- a/include/vptr/virtual_ptr.hpp
+++ b/include/vptr/virtual_ptr.hpp
@@ -35,11 +35,6 @@
 
 #include <CL/sycl.hpp>
 
-#ifdef SYCL_IMPLEMENTATION_ONEAPI
-#if __SYCL_COMPILER_VERSION >= 20220812
-#define NO_CL_NAMESPACE
-#endif
-#endif
 
 #include <cstddef>
 #include <queue>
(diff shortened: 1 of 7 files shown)


include/vptr/virtual_ptr.hpp
```diff
@@ -35,13 +35,21 @@
...
 #include <queue>
 #include <set>
 #include <stdexcept>
 #include <map>
 
+#ifndef NO_CL_NAMESPACE
 namespace cl {
+#endif
 namespace sycl {
 namespace codeplay {
```
DuncanMcBain, line 54: As we'd previously discussed, would you consider changing this to some other namespace, like `vptr` or something? It doesn't really need and likely shouldn't be in the `sycl` namespace at all. Users can remain on the old version if they don't want to change their code, meanwhile we can update SYCL-BLAS when this is merged.

hjabird: The namespace `vptr` is now used instead of `cl::sycl::computecpp`. The tests and samples have been updated to reflect this, and the change still achieves the desired compatability with SYCL-BLAS and DPC++.

ProGTX (approved):

PR #287, opened 2022-10-04
Update vector sample to reflect better practice

DuncanMcBain:
Some misleading comments were removed, and types names updated to match the 2020 specification.

### DIFF
--- a/samples/simple-example-of-vectors.cpp
+++ b/samples/simple-example-of-vectors.cpp
@@ -1,6 +1,6 @@
 /***************************************************************************
  *
- *  Copyright (C) 2016 Codeplay Software Limited
+ *  Copyright (C) Codeplay Software Limited
  *  Licensed under the Apache License, Version 2.0 (the "License");
  *  you may not use this file except in compliance with the License.
  *  You may obtain a copy of the License at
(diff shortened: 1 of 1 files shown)


samples/simple-example-of-vectors.cpp
```diff
@@ -80,9 +72,9 @@ int main() {
        *   Vectors can also be scaled easily using operator overloads */
       cgh.parallel_for<vector_example>(range<3>(4, 4, 4), [=](item<3> item) {
         auto in = ptrA[item.get_linear_id()];
-        float w = in.w();
-        float3 swizzle = in.xyz();
-        float3 scaled = swizzle * w;
+        auto w = in.w();
+        vec<float,3> swizzle = in.xyz();
+        auto scaled = swizzle * w;
```
ProGTX, line 77: Why were these lines changed? It seems like they should have worked.

ProGTX (approved):

samples/simple-example-of-vectors.cpp
DuncanMcBain: They worked fine, but I wanted to show off other ways of spelling the type, as well as `auto` working for per-element access.

PR #290, opened 2022-10-20
Remove binary file

DuncanMcBain:


ProGTX (approved):

samples/CMakeLists.txt
```diff
@@ -32,8 +32,6 @@ add_sample(TARGET builtin-kernel-example SOURCES builtin-kernel-example.cpp)
 add_sample(TARGET example-sycl-application SOURCES example-sycl-application.cpp)
 add_sample(TARGET gaussian-blur SOURCES gaussian-blur.cpp NO_TEST)
 target_include_directories(gaussian-blur PRIVATE ${PROJECT_SOURCE_DIR}/include)
-add_test(NAME gaussian-blur
-  COMMAND gaussian-blur ${CMAKE_CURRENT_SOURCE_DIR}/Lenna.png)
```
alycm, line 36: Removing a test doesn't seem ideal.  All the other samples have a test (via `add_sample`) why shouldn't this one?

DuncanMcBain: It won't run without an image. We could set the CI to ignore failures in this sample specifically but either option still results in the kernel not running.

We could also have it blur random data, I suppose. Or maybe even return success even when there's no image to blur.

